PR #2, opened 2026-03-22
Support configurable schema via schema.yaml

### LINKED ISSUE #1: Support configurable schema via schema.yaml
## Summary

The JSONL entry schema is currently hardcoded — field names, allowed types, SQL columns, CLI args, and FTS columns are all baked into the source. This makes it impossible to adapt the notebook for different research contexts (e.g., tracking `gpu_hours`, `dataset`, `loss`) without editing source code.

## Proposed changes

- Add a `schema.yaml` config file to the notebook directory that defines:
  - Allowed entry types (e.g., observation, decision, result)
  - Custom fields with types (`text`, `integer`, `real`, `list`)
  - Which fields are full-text searchable (`fts: true`)
- Core fields (`id`, `ts`, `writer_id`, `context`, `type`, `content`) remain hardcoded
- `content` is always FTS-indexed
- `--extra key=value` escape hatch for undeclared fields (stored in a JSON blob column)
- `lab-notebook init` generates a default `schema.yaml`
- `lab-notebook rebuild` re-derives SQLite from the current schema config
- CLI args for `emit` are dynamically generated from the schema
- Ship a `schema.example.yaml` reference file in the repo

carbonscott:
## Summary

- Replace hardcoded schema with a `schema.yaml` config file in the notebook directory
- Custom fields become real SQLite columns and dynamic CLI `--field` args
- Fields with `fts: true` are added to the FTS5 virtual table alongside `content`
- `--extra key=value` escape hatch for undeclared fields (stored as JSON blob)
- `lab-notebook init` generates a default `schema.yaml`
- Ship `schema.example.yaml` as a reference file

## Test plan

- [x] All 34 tests pass (`uv run pytest tests/ -v`)
- [ ] Manual smoke test: init, emit with custom fields, sql query on custom columns, search with FTS on custom field, rebuild

Closes #1

### DIFF
--- a/src/lab_notebook/cli.py
+++ b/src/lab_notebook/cli.py
@@ -89,11 +89,16 @@ def load_schema(notebook_dir: Path) -> dict:
     if not isinstance(schema.get("types"), list) or not schema["types"]:
         print("Error: schema.yaml must have a non-empty 'types' list.", file=sys.stderr)
         sys.exit(1)
-    fields = schema.get("fields", {})
+    fields = schema.get("fields") or {}
+    schema["fields"] = fields
     reserved = set(CORE_FIELDS) | {"extra"}
     for name, spec in fields.items():
         if name in reserved:
             print(f"Error: field '{name}' conflicts with a core field.", file=sys.stderr)
+            sys.exit(1)
+        if not isinstance(spec, dict):
+            print(f"Error: field '{name}' must be a mapping (e.g. {{type: text}}), "
+                  f"got '{spec}'.", file=sys.stderr)
             sys.exit(1)
         ftype = spec.get("type")
         if ftype not in VALID_FIELD_TYPES:
@@ -128,7 +133,9 @@ def build_sql(schema: dict) -> SchemaSQL:
         f"CREATE TABLE IF NOT EXISTS entries (\n    "
         + ",\n    ".join(col_defs)
         + "\n);\n"
-        + f"CREATE VIRTUAL TABLE IF NOT EXISTS entries_fts USING fts5({', '.join(fts_cols)});\n"
+        + "CREATE VIRTUAL TABLE IF NOT EXISTS entries_fts USING fts5("
+        + ", ".join(f'"{c}"' for c in fts_cols)
+        + ");\n"
         + "CREATE INDEX IF NOT EXISTS idx_entries_context ON entries(context);\n"
         + "CREATE INDEX IF NOT EXISTS idx_entries_type ON entries(type);\n"
         + "CREATE INDEX IF NOT EXISTS idx_entries_ts ON entries(ts);\n"
@@ -413,6 +420,11 @@ def cmd_emit(args: argparse.Namespace) -> None:
     conn, _, sql = ensure_db(notebook_dir, schema)
     try:
         upsert_entry(conn, flat, sql)
+    except sqlite3.OperationalError:
+        print("Error: SQLite schema is out of date. Run 'lab-notebook rebuild'.",
+              file=sys.stderr)
+        print("(The JSONL entry was saved successfully.)", file=sys.stderr)
+        sys.exit(1)
     finally:
         conn.close()
 
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -159,6 +159,20 @@ class TestSchema:
         with pytest.raises(SystemExit):
             load_schema(tmp_path)
 
+    def test_schema_fields_null(self, notebook):
+        (notebook / "schema.yaml").write_text(
+            "types:\n  - observation\nfields:\n"
+        )
+        schema = load_schema(notebook)
+        assert schema["fields"] == {}
+
+    def test_schema_field_spec_not_dict(self, notebook):
+        (notebook / "schema.yaml").write_text(
+            "types:\n  - observation\nfields:\n  repo: text\n"
+        )
+        with pytest.raises(SystemExit):
+            load_schema(notebook)
+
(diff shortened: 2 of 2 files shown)


carbonscott (commented):
## Review

Five items found, two worth fixing before merge:

1. **Dead code** (`schema_path()`) — remove it
2. **`--extra` collision** — `--extra repo=foo` silently bypasses the declared `repo` field. Should validate.
3. `cmd_emit` duplicates `ensure_db` logic — minor, not blocking
4. Schema parsed twice per `emit` — minor perf, not blocking
5. **README not updated** — still shows old hardcoded examples, no mention of `schema.yaml`

src/lab_notebook/cli.py
```diff
@@ -102,6 +75,87 @@
 """
```
carbonscott: **Bloat #1: `SCHEMA_HELP` is now a lie.**

This 44-line static string describes a hardcoded schema that no longer exists. It mentions `tags` (line 66) which may not be in a custom schema, and does not mention `extra`. The `schema` command just prints this — it is actively misleading now that schemas are configurable.

**Counter-argument:** This is a reference showing _example_ queries, not a live schema dump. But the column list at the top (lines 33-42) claims to be the table structure, which is false for custom schemas. At minimum, the column list should be generated dynamically from `schema.yaml`.

src/lab_notebook/cli.py
```diff
@@ -260,38 +354,61 @@ def cmd_init(args: argparse.Namespace) -> None:
...
+    if args.extra:
+        for item in args.extra:
+            key, _, value = item.partition("=")
+            if not key or not _:
+                print(f"Error: --extra must be key=value, got '{item}'", file=sys.stderr)
+                sys.exit(1)
+            if key in reserved_keys:
+                print(f"Error: --extra key '{key}' conflicts with a declared field. "
+                      f"Use --{key} instead.", file=sys.stderr)
+                sys.exit(1)
```
carbonscott: **Correctness #6: `--extra` values are always strings.**

`--extra gpu_hours=4.5` stores `"4.5"` (a string) in JSONL and the JSON blob. But if `gpu_hours` were a schema field, `cmd_emit` would parse it as `float(4.5)`. The same data has different types depending on how it was declared.

If someone promotes an extra field to a schema field and rebuilds, `rebuild_from_jsonl` does not coerce types — old JSONL entries will have strings where the schema expects numbers.

**Counter-argument:** `--extra` is explicitly the "untyped escape hatch." Users accept the `json_extract` tax and string-only values when they choose `--extra` over declaring a field. Type coercion on rebuild could be a follow-up.

src/lab_notebook/cli.py
```diff
@@ -102,6 +75,87 @@
...
+    col_defs = [
+        "id         TEXT PRIMARY KEY",
+        "ts         TEXT NOT NULL",
+        "writer_id  TEXT NOT NULL",
+        "context    TEXT NOT NULL",
+        "type       TEXT NOT NULL",
+        "content    TEXT NOT NULL",
+    ]
+    for name, spec in fields.items():
+        col_defs.append(f"{name} {TYPE_MAP[spec['type']]}")
```
carbonscott: **Correctness #8: Field names not quoted in generated SQL.**

A schema field named `order`, `group`, `index`, `select`, or any other SQLite reserved word will generate broken SQL:
```sql
CREATE TABLE entries (..., order TEXT, ...)  -- syntax error
```
Fix: wrap field names in double quotes: `"order" TEXT` is valid SQLite.

src/lab_notebook/cli.py
```diff
@@ -134,44 +188,63 @@ def index_path(notebook_dir: Path) -> Path:
...
     conn.commit()
 
 
-def ensure_db(notebook_dir: Path) -> sqlite3.Connection:
+def ensure_db(notebook_dir: Path, schema: dict | None = None) -> tuple[sqlite3.Connection, dict, SchemaSQL]:
+    if schema is None:
+        schema = load_schema(notebook_dir)
+    sql = build_sql(schema)
     dbp = index_path(notebook_dir)
     needs_rebuild = not dbp.exists()
```
carbonscott: **Correctness #9: TOCTOU race in `ensure_db`.**

`needs_rebuild = not dbp.exists()` then `sqlite3.connect(str(dbp))` creates the file. If two processes both see the file as missing, both trigger `rebuild_from_jsonl`.

**Counter-argument:** This is a single-user CLI tool, not a server. Concurrent writes to the same notebook are not a supported use case. The JSONL source of truth is safe regardless; a stale index is fixed by `rebuild`.

tests/test_cli.py
```diff
@@ -19,6 +19,10 @@
     ensure_db,
     entries_dir,
     index_path,
+    load_schema,
+    build_sql,
+    flatten_entry,
+    DEFAULT_SCHEMA_YAML,
```
carbonscott: **Test gap #12: Unused import.** `DEFAULT_SCHEMA_YAML` is imported but never referenced in any test.

**Test gap #11: `make_emit_args` hardcodes field names.** The defaults dict (lines 67-76) and `make_custom_emit_args` (lines 82-92) duplicate schema knowledge. If the default `schema.yaml` changes, tests break with `AttributeError` not clean assertion failures.

**Test gap #10: No `--extra` collision test with `custom_notebook`.** The collision rejection tests (lines 295-301) only use the default schema fixture. There is no test confirming `--extra dataset=x` is rejected when `dataset` is a declared field in the custom schema.

PR #4, opened 2026-03-22
Add bundled schema templates with template utility

### LINKED ISSUE #3: Add bundled schema templates with template utility
## Motivation

The lab-notebook tool supports two distinct use cases via configurable `schema.yaml`:

1. **Research notebook** — sits above repos, captures insights, decisions, dead-ends, and open questions
2. **ML experiment logbook** — lives inside a repo, tracks runs, crashes, hyperparameters, and comparisons

Currently users must write `schema.yaml` from scratch or copy `schema.example.yaml`. We should ship ready-to-use templates and a utility to discover and apply them.

## Changes

- Bundle two schema templates inside the package (`src/lab_notebook/schemas/`)
  - `research-notebook.yaml` — the current default
  - `ml-experiment-log.yaml` — ML experiment tracking with structured hyperparameter fields
- Add `lab-notebook template [name]` subcommand to list/apply templates
- Add `--template` flag to `lab-notebook init`
- Remove `schema.example.yaml` (replaced by bundled templates)
- Update README

carbonscott:
Closes #3

## Summary

- Bundle two schema templates (`research-notebook`, `ml-experiment-log`) inside the package
- Add `lab-notebook template [name]` subcommand to list or apply templates
- Add `--template` flag to `lab-notebook init` for choosing a template at init time
- Remove `schema.example.yaml` (replaced by bundled templates)
- Update README with new commands

## Test plan

- [ ] `lab-notebook template` lists both templates
- [ ] `lab-notebook init /tmp/test --template ml-experiment-log` creates correct schema
- [ ] `lab-notebook init --template` lists templates
- [ ] Default `lab-notebook init` still uses research-notebook template
- [ ] Invalid template name gives clear error
- [ ] `lab-notebook template ml-experiment-log` applies to existing notebook
- [ ] Emit + query works with ML template (structured fields like method, backbone, epochs)
- [ ] All 41 existing tests pass

### DIFF
--- a/src/lab_notebook/cli.py
+++ b/src/lab_notebook/cli.py
@@ -8,7 +8,7 @@ Usage:
     lab-notebook schema
     lab-notebook rebuild
     lab-notebook contexts
-    lab-notebook template [name]
+    lab-notebook template [name] [--force]
 """
 from __future__ import annotations
 
@@ -20,6 +20,7 @@ import sqlite3
 import sys
 from collections import namedtuple
 from datetime import datetime
+from importlib.resources import files
 from pathlib import Path
 
 import yaml
@@ -169,12 +170,16 @@ def build_sql(schema: dict) -> SchemaSQL:
 # Helpers
 # ---------------------------------------------------------------------------
 
-def get_notebook_dir() -> Path:
+def get_notebook_dir(hint: str = "") -> Path:
     d = os.environ.get("LAB_NOTEBOOK_DIR")
     if d:
         return Path(d)
     print("Error: LAB_NOTEBOOK_DIR environment variable is not set.", file=sys.stderr)
-    print("Run 'lab-notebook init' to create a notebook, then source the .env file.", file=sys.stderr)
+    if hint:
+        print(hint, file=sys.stderr)
+    else:
+        print("Run 'lab-notebook init' to create a notebook, then source the .env file.",
+              file=sys.stderr)
     sys.exit(1)
 
 
@@ -302,7 +307,7 @@ def print_table(cursor: sqlite3.Cursor) -> None:
     print(f"\n({len(rows)} row{'s' if len(rows) != 1 else ''})")
 
 
-SCHEMAS_DIR = Path(__file__).parent / "schemas"
+SCHEMAS_DIR = Path(str(files("lab_notebook").joinpath("schemas")))
 DEFAULT_TEMPLATE = "research-notebook"
 
 
@@ -315,7 +320,7 @@ def list_templates() -> list[tuple[str, str]]:
         with open(p) as f:
             first = f.readline().strip()
             if first.startswith("#"):
-                desc = first.lstrip("# ").rstrip(".")
+                desc = first.removeprefix("# ").rstrip(".")
         templates.append((name, desc))
     return templates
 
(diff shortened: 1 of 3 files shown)


src/lab_notebook/cli.py
```diff
@@ -509,6 +548,22 @@ def cmd_contexts(args: argparse.Namespace) -> None:
...
 
+def cmd_template(args: argparse.Namespace) -> None:
+    if not args.name:
+        print_templates()
+        return
+
+    notebook_dir = get_notebook_dir()
+    schema_content = read_template(args.name)
+    sf = notebook_dir / "schema.yaml"
+    sf.write_text(schema_content)
```
carbonscott: **BUG-1 (High): Silent overwrite of customized `schema.yaml`**

`sf.write_text()` runs unconditionally with no confirmation, backup, or `--force` flag. If a user has a carefully customized schema, this destroys it instantly.

**Fix:** Refuse to overwrite unless `--force` is passed. Print a warning showing the existing file.

src/lab_notebook/cli.py
```diff
@@ -342,7 +381,7 @@ def cmd_init(args: argparse.Namespace) -> None:
 
     sf = target / "schema.yaml"
     if not sf.exists():
-        sf.write_text(DEFAULT_SCHEMA_YAML)
+        sf.write_text(read_template(template_name))
```
carbonscott: **BUG-2 (High): `--template` silently ignored when `schema.yaml` exists; output lies**

The `if not sf.exists()` guard skips the write, but line 395 still prints `from template: {template_name}`. The user is told the template was applied when it was not.

**Fix:** When `--template` is explicitly provided and `schema.yaml` already exists, either overwrite it (with warning) or error out. At minimum, do not claim the template was applied.

src/lab_notebook/cli.py
```diff
@@ -301,27 +302,63 @@ def print_table(cursor: sqlite3.Cursor) -> None:
...
+def list_templates() -> list[tuple[str, str]]:
+    """Return [(name, description)] for each .yaml in the bundled schemas dir."""
+    templates = []
+    for p in sorted(SCHEMAS_DIR.glob("*.yaml")):
+        name = p.stem
+        desc = ""
+        with open(p) as f:
+            first = f.readline().strip()
+            if first.startswith("#"):
+                desc = first.lstrip("# ").rstrip(".")
```
carbonscott: **BUG-3 (Low): `lstrip("# ")` strips a character set, not the prefix**

`str.lstrip("# ")` strips any combination of `#` and space from the left, not the literal prefix `"# "`. A comment starting with `# #Special` would yield `Special` instead of `#Special`.

**Fix:** Use `first.removeprefix("# ")` (Python 3.9+, and this project requires 3.11+).

src/lab_notebook/cli.py
```diff
@@ -301,27 +302,63 @@ def print_table(cursor: sqlite3.Cursor) -> None:
...
+        templates.append((name, desc))
+    return templates
+
+
+def get_template_path(name: str) -> Path | None:
+    """Return path to a bundled template, or None if not found."""
+    p = SCHEMAS_DIR / f"{name}.yaml"
+    return p if p.exists() else None
```
carbonscott: **EDGE CASE-1 (Low): Path traversal in template name**

`SCHEMAS_DIR / f"{name}.yaml"` does not validate that the resolved path stays within `SCHEMAS_DIR`. A name like `../../etc/something` could escape the schemas directory.

**Fix:** Add: `if not p.resolve().is_relative_to(SCHEMAS_DIR.resolve()): return None`

PR #6, opened 2026-03-22
Fix: lazy SQLite index rebuild for HPC-safe concurrent writes

### LINKED ISSUE #5: SQLite concurrent writes unsafe on Lustre parallel filesystem
## Problem

Every `lab-notebook emit` call upserts into `index.sqlite`. When multiple agents or batch jobs emit concurrently on Lustre, the unreliable POSIX file locking leads to potential index corruption. The JSONL files are safe (per-writer, append-only), but the shared SQLite index is not.

This affects any HPC workflow where multiple writers call `emit` simultaneously — the typical batch/campaign pattern.

## Proposed fix

- Make `emit` append to JSONL only — never touch SQLite
- Rebuild the SQLite index lazily (on demand) before queries (`sql`, `search`, `contexts`)
- Use mtime-based staleness detection: if any `*.jsonl` or `schema.yaml` is newer than `index.sqlite`, rebuild transparently

This is non-disruptive to users: `emit` gets faster, queries auto-rebuild when needed, and the explicit `rebuild` command still works.

carbonscott:
## Summary

Closes #5

- `emit` no longer writes to `index.sqlite` — only appends to per-writer JSONL files
- SQLite index is rebuilt lazily before queries (`sql`, `search`, `contexts`) using mtime-based staleness detection
- Added `_index_is_stale()` helper: checks if any `*.jsonl` or `schema.yaml` is newer than the index
- Updated tests to reflect the new lazy-rebuild pattern

This makes `lab-notebook` safe for concurrent HPC writes on Lustre, where POSIX file locking is unreliable.

## Test plan

- [x] All 60 existing tests pass
- [ ] Manual: `emit` twice rapidly, confirm both entries appear in `sql` query
- [ ] Manual: confirm `index.sqlite` is not created/modified by `emit`
- [ ] Manual: confirm `sql`/`search`/`contexts` transparently rebuild when index is missing or stale

### DIFF
--- a/src/lab_notebook/cli.py
+++ b/src/lab_notebook/cli.py
@@ -242,12 +242,26 @@ def upsert_entry(conn: sqlite3.Connection, entry: dict, sql: SchemaSQL,
         conn.commit()
 
 
+def _index_is_stale(notebook_dir: Path, dbp: Path) -> bool:
+    """True if any JSONL file or schema.yaml is newer than the index."""
+    idx_mtime = dbp.stat().st_mtime
+    schema_file = notebook_dir / "schema.yaml"
+    if schema_file.exists() and schema_file.stat().st_mtime > idx_mtime:
+        return True
+    edir = entries_dir(notebook_dir)
+    if edir.exists():
+        for f in edir.glob("*.jsonl"):
+            if f.stat().st_mtime > idx_mtime:
+                return True
+    return False
+
+
 def ensure_db(notebook_dir: Path, schema: dict | None = None) -> tuple[sqlite3.Connection, dict, SchemaSQL]:
     if schema is None:
         schema = load_schema(notebook_dir)
     sql = build_sql(schema)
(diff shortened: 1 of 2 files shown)


carbonscott (commented):
Adversarial review: the write-path fix is correct, but there are a few issues worth addressing before this is HPC-safe end-to-end.

**Inconsistency (low): `cmd_rebuild` (line 544) still writes SQLite in-place**

`cmd_rebuild` deletes `index.sqlite` then creates and populates a new one in-place. If a query command (`sql`, `search`, `contexts`) runs mid-rebuild, it will hit a missing or partially-populated index. The same atomic-rename fix suggested for `ensure_db` should apply here: rebuild into a temp file, then rename into place.

carbonscott (commented):
Adversarial review of second commit (cfef7c5). All 5 original findings are addressed. Remaining items are minor — no blocking issues.

PR #13, opened 2026-04-03
Make project-local init the default, remove --local flag

### LINKED ISSUE #12: Make lab-notebook init always create .lnb/ + .lnb.env (remove --local flag)
## Problem

`lab-notebook init` currently has two modes:
- **Default**: initializes notebook in CWD with `.env` inside it
- **`--local`**: creates `.lnb/` in CWD + `.lnb.env` for discovery

The `--local` pattern is the primary workflow (used by the `/lnb` skill), but it requires an extra flag. Running bare `lab-notebook init` doesn't create a `.lnb` folder, which is confusing.

## Solution

Simplify to one behavior — always project-local:
- `lab-notebook init` → creates `.lnb/` in CWD + `.lnb.env` in CWD
- `lab-notebook init <path>` → creates notebook at `<path>` + `.lnb.env` in CWD
- Remove `--local` flag entirely
- Remove the old non-local code path (`.env` inside notebook dir)

## Files to change
- `src/lab_notebook/cli.py` — `cmd_init`, argparse setup, `get_notebook_dir` hint
- `tests/test_cli.py` — update fixtures and init tests
- `README.md` — update Quick Start, Commands, Data Layout

carbonscott:
## Summary

Closes #12

- `lab-notebook init` now always creates `.lnb/` + `.lnb.env` in the current directory (previously required `--local`)
- `lab-notebook init <path>` creates the notebook at `<path>` + `.lnb.env` in CWD
- Removed `--local` flag and the old non-local code path (`.env` inside notebook dir)
- Updated tests and README to match

## Test plan

- [x] All 81 tests pass (`uv run pytest tests/test_cli.py -v`)
- [x] Smoke test: `lab-notebook init` creates `.lnb/` with `entries/`, `artifacts/`, `schema.yaml`, `.gitignore`
- [x] Smoke test: `.lnb.env` written to CWD with correct `LAB_NOTEBOOK_DIR`
- [ ] Verify `source .lnb.env && lab-notebook schema` works end-to-end

### DIFF
--- a/src/lab_notebook/cli.py
+++ b/src/lab_notebook/cli.py
@@ -490,7 +490,7 @@ def cmd_init(args: argparse.Namespace) -> None:
     print(f"  LAB_NOTEBOOK_DIR={target}")
     print(f"\nConsider adding to .gitignore:")
     print(f"  {LNB_ENV_FILE}")
-    print(f"  .lnb/")
+    print(f"  {target.name}/")
 
 
 def cmd_emit(args: argparse.Namespace) -> None:
@@ -656,7 +656,7 @@ def main() -> None:
     # -- init --
     p_init = sub.add_parser("init", help="Initialize a notebook directory")
     p_init.add_argument("path", nargs="?", default=None,
-                        help="Notebook directory (default: .lnb in current directory)")
+                        help="Notebook root directory (default: .lnb in current directory)")
     p_init.add_argument("--template", nargs="?", const="", default=None,
                         help="Schema template to use (omit value to list available templates)")
     p_init.set_defaults(func=cmd_init)
--- a/src/lab_notebook/cli.py
+++ b/src/lab_notebook/cli.py
@@ -232,7 +232,7 @@ def get_notebook_dir(hint: str = "") -> Path:
     if hint:
         print(hint, file=sys.stderr)
     else:
-        print("Run 'lab-notebook init --local' to set up a project notebook,\n"
+        print("Run 'lab-notebook init' to set up a project notebook,\n"
               "or set $LAB_NOTEBOOK_DIR in your shell profile.",
               file=sys.stderr)
     sys.exit(1)
@@ -443,21 +443,8 @@ def cmd_init(args: argparse.Namespace) -> None:
         print_templates()
         return
 
-    local = getattr(args, "local", False)
-
-    if local:
-        # --local: default notebook path is .lnb in CWD
-        target = Path(args.path or ".lnb").resolve()
-        target.mkdir(parents=True, exist_ok=True)
-    else:
-        target = Path(args.path or ".").resolve()
-
-    if not target.exists():
-        print(f"Error: directory does not exist: {target}", file=sys.stderr)
-        sys.exit(1)
-    if not target.is_dir():
-        print(f"Error: not a directory: {target}", file=sys.stderr)
-        sys.exit(1)
+    target = Path(args.path or ".lnb").resolve()
+    target.mkdir(parents=True, exist_ok=True)
 
     template_name = getattr(args, "template", None) or DEFAULT_TEMPLATE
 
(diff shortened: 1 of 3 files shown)


tests/test_cli.py
```diff
@@ -145,14 +147,18 @@ def test_init_cwd(self, tmp_path, monkeypatch):
...
-        args = argparse.Namespace(path=str(tmp_path / "does-not-exist"), template=None)
-        with pytest.raises(SystemExit):
-            cmd_init(args)
+    def test_init_auto_creates_dir(self, tmp_path, monkeypatch):
+        monkeypatch.chdir(tmp_path)
+        target = tmp_path / "new-notebook"
+        args = argparse.Namespace(path=str(target), template=None)
+        cmd_init(args)
+        assert target.is_dir()
+        assert (target / "entries").is_dir()
```
carbonscott, line 161: Nit: `test_init_auto_creates_dir` checks that the directory and `entries/` subdirectory are created, but does not assert that `.lnb.env` was written to `tmp_path`. All sibling tests that call `cmd_init` verify the env file; adding `assert (tmp_path / LNB_ENV_FILE).exists()` here would keep coverage consistent.

src/lab_notebook/cli.py
```diff
@@ -485,40 +472,25 @@ def cmd_init(args: argparse.Namespace) -> None:
...
+    print(f"Initialized lab notebook in {target}")
+    print(f"  entries/       per-writer JSONL files")
+    print(f"  artifacts/     files referenced via --artifacts")
+    print(f"  schema.yaml    {schema_msg}")
+    print(f"  .gitignore     ignores index.sqlite")
+    print(f"\nCreated {lnb_env.name} in {lnb_env.parent}")
+    print(f"  LAB_NOTEBOOK_DIR={target}")
+    print(f"\nConsider adding to .gitignore:")
+    print(f"  {LNB_ENV_FILE}")
+    print(f"  .lnb/")
```
carbonscott: Nit: The hint `"  .lnb/"` is hard-coded. If the user ran `lab-notebook init custom-name`, the hint will still suggest `.lnb/`, which is misleading. Consider printing `f"  {target.name}/"` or deriving the name from `target` so the suggestion matches what was actually created.

src/lab_notebook/cli.py
```diff
@@ -684,9 +656,7 @@ def main() -> None:
     # -- init --
     p_init = sub.add_parser("init", help="Initialize a notebook directory")
     p_init.add_argument("path", nargs="?", default=None,
-                        help="Directory to initialize (default: current directory)")
-    p_init.add_argument("--local", action="store_true",
-                        help="Create .lnb.env in CWD for project-local notebook (default path: .lnb)")
+                        help="Notebook directory (default: .lnb in current directory)")
```
carbonscott: Consider: The help string `"Notebook directory (default: .lnb in current directory)"` could also note that when a path is given it is used as-is (not nested inside `.lnb/`), e.g. `"Notebook directory; created if absent (default: .lnb)"`. This prevents users from expecting `lab-notebook init foo` to create `foo/.lnb/`.

carbonscott (commented):
## Review Summary

This PR cleanly simplifies `lab-notebook init` by making project-local behavior (`.lnb/` + `.lnb.env`) the default and removing the old `--local` flag. The change reduces user-facing complexity without losing any capability, the tests are comprehensive and well-updated, and the README accurately reflects the new UX. Overall the PR is in good shape — all findings below are non-blocking nits.

## Key Findings

### Functionality

- **`lab-notebook init <path>` always uses that path verbatim as the notebook root.** With the old `--local` flag the notebook always lived inside `.lnb/` and the env file pointed there; now `lab-notebook init my-notebook` creates `my-notebook/` (not `my-notebook/.lnb/`). This is arguably the intended behavior but it means `.lnb.env` will say `LAB_NOTEBOOK_DIR=/abs/path/my-notebook`, which is a flat directory — not nested under `.lnb`. The README example and the `test_init_custom_path` test confirm this is deliberate, and that's fine, but it could surprise users who assume `init <path>` always places things in `<path>/.lnb`. A one-line note in the help text would prevent confusion.

- **`lab-notebook init` silently overwrites an existing `.lnb.env`.** The warning is printed to stderr, which is good, but there is no prompt or `--force` guard. The previous code had the same behavior, so this is not a regression — just worth noting.

### Tests

- `test_init_auto_creates_dir` (line 155 in tests/test_cli.py) does not assert that `.lnb.env` was written to `tmp_path`. Every other test that calls `cmd_init` with a non-default path checks for the env file; this one does not. It would make the test more complete.

### Documentation

- The `init` command help text (`"Notebook directory (default: .lnb in current directory)"`) does not clarify that when `path` is given it is used directly as the notebook directory (not nested inside `.lnb`). A short clarification would prevent user confusion.

## Positives

- The deletion of the old non-local code path is surgical and leaves no dead code.
- `monkeypatch.chdir(tmp_path)` added to every relevant test ensures `.lnb.env` is written to the right temp directory and tests remain isolated — good defensive testing.
- Renaming `TestInitLocal` to `TestInitDefault` accurately reflects the new semantics.
- The `test_init_auto_creates_dir` replacement for `test_init_nonexistent_dir` correctly inverts the old behavior check to match the new `parents=True` semantics.

PR #14, opened 2026-04-03
Always create .lnb subfolder even with explicit path

carbonscott:
## Summary

Follow-up to #13. `lab-notebook init <path>` previously used `<path>` as the
notebook root directly. Now it always creates `<path>/.lnb/`, making the
behavior consistent whether or not a path is given.

- `lab-notebook init` → `CWD/.lnb/` (unchanged)
- `lab-notebook init /tmp/foo` → `/tmp/foo/.lnb/` (fixed)

One-line change in `cmd_init`, rest is test updates.

## Test plan

- [x] All 81 tests pass
- [ ] `lab-notebook init /tmp/test` creates `/tmp/test/.lnb/` with entries/, artifacts/, schema.yaml
- [ ] `source .lnb.env && lab-notebook schema` works from inside the project dir

### DIFF
--- a/src/lab_notebook/cli.py
+++ b/src/lab_notebook/cli.py
@@ -443,7 +443,7 @@ def cmd_init(args: argparse.Namespace) -> None:
         print_templates()
         return
 
-    target = (Path(args.path) / ".lnb" if args.path else Path(".lnb")).resolve()
+    target = (Path(args.path or ".") / ".lnb").resolve()
     target.mkdir(parents=True, exist_ok=True)
 
     template_name = getattr(args, "template", None) or DEFAULT_TEMPLATE
--- a/src/lab_notebook/cli.py
+++ b/src/lab_notebook/cli.py
@@ -443,7 +443,7 @@ def cmd_init(args: argparse.Namespace) -> None:
         print_templates()
         return
 
-    target = Path(args.path or ".lnb").resolve()
+    target = (Path(args.path) / ".lnb" if args.path else Path(".lnb")).resolve()
     target.mkdir(parents=True, exist_ok=True)
 
     template_name = getattr(args, "template", None) or DEFAULT_TEMPLATE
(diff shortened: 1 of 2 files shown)


src/lab_notebook/cli.py
```diff
@@ -443,7 +443,7 @@ def cmd_init(args: argparse.Namespace) -> None:
         print_templates()
         return
 
-    target = Path(args.path or ".lnb").resolve()
+    target = (Path(args.path) / ".lnb" if args.path else Path(".lnb")).resolve()
```
carbonscott: Nit: The two `Path(...)` constructions can be collapsed into one:

```python
target = (Path(args.path or ".") / ".lnb").resolve()
```

`args.path or "."` evaluates to `"."` when `args.path` is falsy, so `Path(".") / ".lnb"` resolves to the same `CWD/.lnb` as the current code. This removes the repeated `Path(...)` call and makes both branches structurally identical. Purely optional.

carbonscott (commented):
## Review Summary

Clean, well-scoped follow-up to #13. The production change is a single line and the intent is immediately clear. Test updates are thorough — every fixture and assertion that relied on the old layout is updated consistently, and the renamed variable (`nb_dir` vs `target`) improves clarity at the call sites.

### Key Findings

**Functionality** — Correct. `(Path(args.path) / ".lnb" if args.path else Path(".lnb")).resolve()` handles both the explicit-path and no-path cases. The `.resolve()` call properly anchors relative paths (e.g. `"my-project"`) to the cwd at call time, which is the right behaviour.

**Complexity** — The one-liner is readable enough, but see the inline nit below for an alternative that removes the repeated `Path(...)` construction.

**Tests** — Comprehensive. Every affected fixture (`notebook`, `custom_notebook`) and test class (`TestInit`, `TestInitTemplate`, `TestInitLocalEnv`) is updated. The 81-test pass count noted in the PR description gives good confidence.

**PR Description** — Good. The before/after bullet points make the behavioural change concrete and easy to verify manually. The unchecked items in the test plan are honest about what still needs manual verification.

### Good things

Renaming `target` -> `nb_dir` in the test fixtures is a small but meaningful improvement — `nb_dir` makes it unambiguous that the variable refers to the `.lnb/` directory, not the parent project directory.

Overall: this definitively improves code health. LGTM.

PR #16, opened 2026-04-15
feat: add --template-path to init for external schema files

### LINKED ISSUE #15: Extend `init --template` to accept a file path
## Motivation

Currently `lab-notebook init --template NAME` resolves `NAME` only against bundled templates in `src/lab_notebook/schemas/` (see `cli.py:401-417`, `get_template_path`). Projects that want to ship a custom schema have to do the init → overwrite `schema.yaml` → `rebuild` dance manually.

Concrete case: [ralph-wiggum-lnb](https://github.com/carbonscott/ralph-wiggum-lnb) ships a `coding-dev.yaml` schema and its `ensure_notebook` helper currently has to:

1. `lab-notebook init` (writes default `research-notebook` schema)
2. `cp coding-dev.yaml .lnb/schema.yaml` (overwrite)
3. `lab-notebook rebuild` (re-sync SQLite type vocabulary)

…plus a fallback warning branch if the file is missing. All of this would collapse into a single `lab-notebook init --template ./coding-dev.yaml` if `--template` accepted a path.

## Proposal

Extend `--template` to accept an **external file path** as a fallback when the argument is not a bundled template name:

- If the argument resolves to an existing file on disk (absolute or relative), load the schema from that file.
- Otherwise, fall back to the current behavior: look up `SCHEMAS_DIR / f"{name}.yaml"` and error with "unknown template" if not found.

This keeps the bundled-template UX intact (`--template coding-dev` still works if/when bundled) while letting downstream projects ship their own schemas without post-init surgery.

## Sketch

In `read_template(name)` / `get_template_path(name)` (cli.py:401-417), add a path-check branch before the bundled lookup:

```python
def get_template_path(name: str) -> Path | None:
    # External file path fallback
    p = Path(name)
    if p.is_file() and p.suffix in (".yaml", ".yml"):
        return p.resolve()
    # Bundled template lookup (existing behavior)
    bundled = SCHEMAS_DIR / f"{name}.yaml"
    if not bundled.resolve().is_relative_to(SCHEMAS_DIR.resolve()):
        return None
    return bundled if bundled.exists() else None
```

Disambiguation rule: the file-path branch wins only when the argument ends in `.yaml`/`.yml` or contains a path separator. Bare names like `coding-dev` continue to resolve against the bundled set. This avoids surprising users who expect bare names to be bundled.

## Alternatives considered

- **Bundle `coding-dev` upstream.** Cleaner for the ralph-wiggum-lnb case specifically, but does not generalize — every downstream project with a custom schema would need a PR here. The path-fallback approach solves the class of problem.
- **Status quo (manual overwrite + rebuild).** Works but duplicates logic across every downstream harness.

## Scope

Small change, contained to `get_template_path` / `read_template` in `cli.py`. Print-templates listing and the default-template behavior stay as-is. Happy to send a PR if the direction looks right.

carbonscott:
## Summary

- Adds `--template-path PATH` to `lab-notebook init`, mutually exclusive with `--template NAME`, so downstream projects can initialize a notebook from their own schema file in one command.
- `--template` keeps resolving against bundled `SCHEMAS_DIR`; `--template-path` loads a schema YAML from any file on disk.
- Schema validation still lives in `load_schema`, so a malformed custom schema errors cleanly on first use — no new validation code.

## Design note

Issue #15 originally sketched overloading `--template` with path-sniffing (".yaml" suffix → treat as path, else bundled lookup). We rejected that in favor of a separate flag because:

- No magic suffix heuristic, no risk of a stray `coding-dev` file shadowing a bundled name.
- Error messages stay unambiguous: "unknown template" and "file not found" are distinct failures.
- Matches the standard Unix `--foo` / `--foo-file` convention.

Mutex is enforced via argparse `add_mutually_exclusive_group()` — first use in this codebase.

## Test plan

- [x] `pytest tests/test_cli.py` — all 89 tests pass (9 new)
- [x] Smoke test: `init --template-path ./custom.yaml` in a scratch dir, then `emit` + `sql` against the custom schema
- [x] Mutex check: `init --template foo --template-path bar.yaml` → argparse error, exit 2
- [x] Missing-file check: `init --template-path ./nope.yaml` → `Error: template path not found or not a file: ...`, exit 1
- [x] README updated (init command + Schema Configuration section)

Closes #15.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/lab_notebook/cli.py
+++ b/src/lab_notebook/cli.py
@@ -417,6 +417,21 @@ def read_template(name: str) -> str:
     return p.read_text()
 
 
+def read_template_from_path(path: str) -> str:
+    """Read a schema template from an external file path, or exit with error."""
+    p = Path(path)
+    if not p.is_file():
+        print(f"Error: template path not found or not a file: {path}",
+              file=sys.stderr)
+        sys.exit(1)
+    try:
+        return p.read_text()
+    except OSError as e:
+        print(f"Error: failed to read template at {path}: {e}",
+              file=sys.stderr)
+        sys.exit(1)
+
+
 def print_templates() -> None:
     """Print available templates to stdout."""
     if not SCHEMAS_DIR.is_dir():
@@ -446,7 +461,21 @@ def cmd_init(args: argparse.Namespace) -> None:
     target = (Path(args.path or ".") / ".lnb").resolve()
     target.mkdir(parents=True, exist_ok=True)
 
-    template_name = getattr(args, "template", None) or DEFAULT_TEMPLATE
+    template_path = getattr(args, "template_path", None)
+    template_name = getattr(args, "template", None)
+
+    if template_path is not None:
+        schema_text = read_template_from_path(template_path)
+        schema_source = f"from path: {template_path}"
+        explicit = True
+    elif template_name is not None:
+        schema_text = read_template(template_name)
+        schema_source = f"from template: {template_name}"
+        explicit = True
+    else:
+        schema_text = read_template(DEFAULT_TEMPLATE)
+        schema_source = f"from template: {DEFAULT_TEMPLATE}"
+        explicit = False
 
     edir = target / "entries"
     edir.mkdir(exist_ok=True)
@@ -460,13 +489,12 @@ def cmd_init(args: argparse.Namespace) -> None:
             f.write("index.sqlite\n")
 
     sf = target / "schema.yaml"
-    template_explicit = getattr(args, "template", None) is not None
     if not sf.exists():
-        sf.write_text(read_template(template_name))
-        schema_msg = f"from template: {template_name}"
-    elif template_explicit:
-        sf.write_text(read_template(template_name))
-        schema_msg = f"from template: {template_name} (overwritten)"
+        sf.write_text(schema_text)
+        schema_msg = schema_source
+    elif explicit:
+        sf.write_text(schema_text)
+        schema_msg = f"{schema_source} (overwritten)"
     else:
         schema_msg = "already exists (kept)"
 
@@ -657,8 +685,11 @@ def main() -> None:
     p_init = sub.add_parser("init", help="Initialize a notebook directory")
     p_init.add_argument("path", nargs="?", default=None,
                         help="Notebook root directory (default: .lnb in current directory)")
-    p_init.add_argument("--template", nargs="?", const="", default=None,
-                        help="Schema template to use (omit value to list available templates)")
+    tgroup = p_init.add_mutually_exclusive_group()
+    tgroup.add_argument("--template", nargs="?", const="", default=None,
+                        help="Bundled schema template (omit value to list available templates)")
+    tgroup.add_argument("--template-path", default=None, metavar="PATH",
+                        help="Load schema from a YAML file on disk")
     p_init.set_defaults(func=cmd_init)
 
     # -- emit --
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -31,8 +31,10 @@ from lab_notebook.cli import (
     load_schema,
     build_sql,
     flatten_entry,
+    main,
     print_templates,
     read_template,
+    read_template_from_path,
 )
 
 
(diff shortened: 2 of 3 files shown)


src/lab_notebook/cli.py
```diff
@@ -417,6 +417,21 @@ def read_template(name: str) -> str:
     return p.read_text()
 
 
+def read_template_from_path(path: str) -> str:
+    """Read a schema template from an external file path, or exit with error."""
+    p = Path(path)
+    if not p.is_file():
```
carbonscott, line 423: FYI: Using `Path.is_file()` as the single gate here is nice -- it returns False for both non-existent paths and directories, so the one error message "not found or not a file" correctly handles both cases without a second `is_dir()` branch. The extra `OSError` guard around `read_text()` is also a small improvement over `read_template` above, which would propagate a raw `OSError` on e.g. permission denied. If you want perfect symmetry you could lift the same pattern up into `read_template` in a follow-up, but it's not needed for this PR.

src/lab_notebook/cli.py
```diff
@@ -446,7 +461,21 @@ def cmd_init(args: argparse.Namespace) -> None:
...
+        schema_source = f"from path: {template_path}"
+        explicit = True
+    elif template_name is not None:
+        schema_text = read_template(template_name)
+        schema_source = f"from template: {template_name}"
+        explicit = True
+    else:
+        schema_text = read_template(DEFAULT_TEMPLATE)
+        schema_source = f"from template: {DEFAULT_TEMPLATE}"
+        explicit = False
```
carbonscott, line 478: Nice refactor. The `(schema_text, schema_source, explicit)` triple computed here cleanly replaces the old `template_name` + `template_explicit` pair, and the downstream `if not sf.exists() / elif explicit / else` block below is unchanged in meaning. Traced it against the pre-PR version and the three cases map 1:1.

tests/test_cli.py
```diff
@@ -791,6 +811,67 @@ def test_init_output_reflects_reality(self, tmp_path, monkeypatch, capsys):
...
+        monkeypatch.chdir(tmp_path)
+        custom = tmp_path / "custom.yaml"
+        custom.write_text("types:\n  - note\nfields: {}\n")
+        monkeypatch.setattr(
+            "sys.argv",
+            ["lab-notebook", "init", "--template", "research-notebook",
+             "--template-path", str(custom)],
+        )
+        with pytest.raises(SystemExit):
+            main()
```
carbonscott, line 873: FYI: Good call exercising the mutex through `main()` + `monkeypatch.setattr("sys.argv", ...)` rather than constructing an `argparse.Namespace` by hand. A Namespace-level test would silently pass even if the `add_mutually_exclusive_group()` wiring were removed -- this one would actually fail, which is what you want from a regression test for argparse behavior.

carbonscott (commented):
## Review Summary

**Overall:** Clean, well-scoped change. The argparse mutex group + separate `--template-path` is the right call over suffix-sniffing `--template`, the refactor preserves the old three-way branch semantics exactly, and the tests cover the important paths. LGTM (switching to COMMENT event since GitHub does not allow self-approval).

### Key Findings

**Design**
- The new-flag-vs-overloading decision is well motivated in the PR body and matches Unix conventions (`--foo` / `--foo-file`). Good to see argparse `add_mutually_exclusive_group()` used for the enforcement rather than hand-rolled checks in `cmd_init`.
- `read_template_from_path` mirrors `read_template`'s exit-on-error contract, so callers downstream of both helpers behave uniformly.

**Functionality**
- Traced the `cmd_init` refactor line by line against the pre-PR code: the three branches (fresh write / explicit overwrite / implicit keep) still map 1:1, and the rename `template_explicit` -> `explicit` is just a local. Existing `TestInitTemplate` cases that build `argparse.Namespace(path=..., template=...)` without `template_path` keep working because the code uses `getattr(args, "template_path", None)`.
- Error message "template path not found or not a file" correctly covers both missing and directory-as-path cases in one check (since `Path.is_file()` is False for both). Nice.

**Tests**
- Coverage for the new flag is solid: valid file, missing file, directory, overwrite-existing-schema, output message, and a real argparse-level mutex test via `main()` + monkeypatched `sys.argv` (not just a hand-built Namespace). That mutex test is the right shape -- it would catch regressions in the argparse wiring, whereas a Namespace-only test would not.
- Optional gap: the PR body says "a malformed custom schema errors cleanly on first use" via `load_schema`, but no test in this PR exercises that path end-to-end (init with a syntactically broken YAML, then `emit` / `sql`). `load_schema` already has its own validation tests so this is a Nit, not blocking -- worth adding one integration test later to lock in the contract the PR description makes.

**Complexity**
- `cmd_init` is now a bit longer but easier to follow: one spot decides `(schema_text, schema_source, explicit)` and the downstream write logic stays unchanged. No over-engineering.

**Naming / Style / Docs**
- `schema_source` vs `schema_msg` read clearly in context. README updates in both the intro blurb and the `init` command section match the flag's behavior.

Nothing blocking. 2 inline comments, both positive/FYI.

PR #18, opened 2026-04-17
fix: honor explicit $LAB_NOTEBOOK_DIR over .lnb.env auto-discovery

### LINKED ISSUE #17: lab-notebook sql returned entries that don't exist in JSONL (stale index?)
## What happened

While smoke-testing an unrelated project, I queried a fresh notebook with `lab-notebook sql` and got back entries that didn't exist in the notebook's `entries/*.jsonl` files.

Specifically:
- Sandbox `.lnb/` was freshly created by `lab-notebook init --template-path .../coding-dev.yaml`
- The run produced entries in `.lnb/entries/cwang31.jsonl` with `ts` values from today and `type` values from the coding-dev schema (`start`, `impl`, `test`, `done`)
- `LAB_NOTEBOOK_DIR=/path/to/sandbox/.lnb lab-notebook sql "SELECT ts, type, context, tags, substr(content,1,80) FROM entries ORDER BY ts"` returned entries with `ts` values from **yesterday** and `type=milestone` (a type that only exists in the `research-notebook` template, not `coding-dev`)

So the SQL output had **zero overlap** with the JSONL contents — as if the query hit a different, pre-existing index rather than the one matching the sandbox's JSONL.

## What I expected

`lab-notebook sql` to reflect the contents of `$LAB_NOTEBOOK_DIR/entries/*.jsonl`. At minimum, to never return rows that don't exist in the JSONL.

## Possible causes I can think of

1. **Stale `index.sqlite`**: The [Commands section of the README](https://github.com/carbonscott/lab-notebook#commands) says `sql` "Auto-rebuilds if `index.sqlite` is missing." It doesn't auto-rebuild if the index is *present but stale*. If something populated `index.sqlite` before the fresh entries were appended (unlikely for a brand-new `.lnb/`, but possible if `init` leaves a seeded index?), the query would miss new rows.
2. **`LAB_NOTEBOOK_DIR` env ignored in some path**: Maybe the CLI auto-discovers a closer `.lnb.env` / upward-walks and finds a different notebook before honoring the env var? Would be surprising but would explain the "different notebook entirely" observation.
3. **`init` leaves a `research-notebook`-schema `index.sqlite`** even when `--template-path` points at a different YAML — and subsequent `emit` calls update the JSONL but don't catch up the schema mismatch in the index?

I haven't dug into the code, so any of these could be off-base.

## Workaround

`lab-notebook rebuild` before `sql` makes the problem go away. Output matches the JSONL as expected.

## Repro (rough)

1. `mkdir /tmp/lnb-repro && cd /tmp/lnb-repro`
2. `lab-notebook init --template-path /path/to/coding-dev.yaml`
3. Populate with a few `lab-notebook emit ...` calls (use coding-dev types like `impl`, `done`)
4. `LAB_NOTEBOOK_DIR=/tmp/lnb-repro/.lnb lab-notebook sql "SELECT ts, type FROM entries ORDER BY ts"`

If the SQL output doesn't match `entries/*.jsonl`, repro confirmed.

## Ask

Either (a) auto-detect stale index (compare JSONL mtimes vs `index.sqlite` mtime, auto-rebuild if out of date), or (b) print a clear warning in `sql` output when staleness is suspected, or (c) document the `rebuild`-after-external-writes requirement more prominently in the README. (a) would be the most user-friendly.

Happy to verify with more repro detail if useful.

carbonscott:
## Summary

- Fixes #17. `get_notebook_dir()` used to resolve `.lnb.env` (walked upward from CWD) before `$LAB_NOTEBOOK_DIR`, so `LAB_NOTEBOOK_DIR=/sandbox/.lnb lab-notebook sql ...` silently ran against a different notebook whenever any parent directory had a `.lnb.env`.
- Flips precedence to the standard Unix order: explicit env var wins, `.lnb.env` walk-up only fills in when the env var is unset. Empty-string env var is treated as unset (falls through).
- Same swap applied to the best-effort schema lookup in `cmd_emit` so `--<field>` argparse options stay consistent with the resolved notebook.

## Test plan

- [x] `uv run pytest tests/test_cli.py -v` — 90 passed, including renamed `test_get_notebook_dir_env_var_precedence` (flipped assertion) and new `test_get_notebook_dir_empty_env_var_falls_through`.
- [x] Manual repro from issue #17: outer notebook at `/tmp/lnb-outer/.lnb`, inner sandbox at `/tmp/lnb-outer/sub/sandbox/.lnb` with a distinct entry. From `/tmp/lnb-outer/sub/` (where walk-up hits the outer `.lnb.env`):
  - `LAB_NOTEBOOK_DIR=.../sandbox/.lnb lab-notebook sql "..."` → returns sandbox entry. ✅
  - Without env var → falls through to walk-up, returns outer notebook. ✅
- [ ] README note on precedence renders correctly on GitHub.

### DIFF
--- a/src/lab_notebook/cli.py
+++ b/src/lab_notebook/cli.py
@@ -217,23 +217,23 @@ def _parse_lnb_env(env_file: Path) -> str | None:
 
 
 def get_notebook_dir(hint: str = "") -> Path:
-    # 1. Check for .lnb.env (walk up from CWD)
+    # 1. Explicit $LAB_NOTEBOOK_DIR wins (standard Unix precedence)
+    d = os.environ.get("LAB_NOTEBOOK_DIR")
+    if d:
+        return Path(d)
+    # 2. Fall back to nearest .lnb.env walking up from CWD
     env_file = _find_lnb_env()
     if env_file:
         val = _parse_lnb_env(env_file)
         if val:
             return Path(val)
-    # 2. Fall back to environment variable
-    d = os.environ.get("LAB_NOTEBOOK_DIR")
-    if d:
-        return Path(d)
     # 3. Error
     print("Error: LAB_NOTEBOOK_DIR is not set and no .lnb.env found.", file=sys.stderr)
     if hint:
         print(hint, file=sys.stderr)
     else:
-        print("Run 'lab-notebook init' to set up a project notebook,\n"
-              "or set $LAB_NOTEBOOK_DIR in your shell profile.",
+        print("Set $LAB_NOTEBOOK_DIR, or run 'lab-notebook init' to create\n"
+              "a project-local notebook (writes .lnb.env in the current directory).",
               file=sys.stderr)
     sys.exit(1)
 
@@ -706,8 +706,10 @@ def main() -> None:
     # Best-effort: don't crash arg parsing if the path is stale or schema is bad.
     _parsed_schema = None
     try:
-        env_file = _find_lnb_env()
-        notebook_env = (_parse_lnb_env(env_file) if env_file else None) or os.environ.get("LAB_NOTEBOOK_DIR")
+        notebook_env = os.environ.get("LAB_NOTEBOOK_DIR")
+        if not notebook_env:
+            env_file = _find_lnb_env()
+            notebook_env = _parse_lnb_env(env_file) if env_file else None
         if notebook_env:
             nb_dir = Path(notebook_env)
             sf = nb_dir / "schema.yaml"
--- a/README.md
+++ b/README.md
@@ -53,6 +53,19 @@ lab-notebook search "masking"
 | `LAB_NOTEBOOK_DIR` | Yes | — | Path to notebook data directory |
 | `LAB_NOTEBOOK_WRITER` | No | `$USER` | Writer ID for entries |
 
+### Notebook discovery precedence
+
+When the CLI resolves which notebook to use, it checks sources in this order
+(first match wins):
+
+1. `$LAB_NOTEBOOK_DIR` environment variable (if set and non-empty)
+2. Nearest `.lnb.env` walking up from the current directory (closest wins,
+   stops at `$HOME` or `/`)
+
+Explicit `LAB_NOTEBOOK_DIR=... lab-notebook ...` always wins, so one-shot
+overrides work as expected. For normal project use, `source .lnb.env` after
+`cd`ing into a project so the env var is set for your shell.
+
 ## Schema Configuration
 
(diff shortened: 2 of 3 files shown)


src/lab_notebook/cli.py
```diff
@@ -706,8 +706,10 @@ def main() -> None:
     # Best-effort: don't crash arg parsing if the path is stale or schema is bad.
     _parsed_schema = None
     try:
-        env_file = _find_lnb_env()
-        notebook_env = (_parse_lnb_env(env_file) if env_file else None) or os.environ.get("LAB_NOTEBOOK_DIR")
+        notebook_env = os.environ.get("LAB_NOTEBOOK_DIR")
+        if not notebook_env:
+            env_file = _find_lnb_env()
+            notebook_env = _parse_lnb_env(env_file) if env_file else None
```
carbonscott, line 712: Optional / FYI: This block now mirrors the precedence logic in `get_notebook_dir()` exactly. Not worth fixing in this PR, but if a future change adds a third resolution site, extracting something like `_resolve_notebook_dir() -> Path | None` would keep the rule in one place and prevent the two copies from drifting. Happy to see the logic become more straightforward here regardless.

carbonscott (commented):
## Review Summary

**Overall:** Clean, well-scoped bug fix. The precedence flip is the right call, the change is minimal, and tests + README are updated in the same PR. Would be `APPROVE` — posting as `COMMENT` because GitHub blocks self-approval on author's own PR.

### Key Findings

**Design / Functionality**
- Precedence now matches standard Unix convention (explicit env var beats config file). Fixes #17 exactly as described.
- Treating an empty-string `LAB_NOTEBOOK_DIR` as unset is the right call — matches typical shell semantics and avoids a surprising error when a user unsets via `LAB_NOTEBOOK_DIR=`.

**Tests**
- The renamed `test_get_notebook_dir_env_var_precedence` correctly flips the assertion, and the new `test_get_notebook_dir_empty_env_var_falls_through` pins down the empty-string behavior. Good coverage for a precedence bug.

**Complexity**
- The `cmd_emit` schema-lookup block is actually easier to read after the change — the explicit two-step lookup beats the previous compound `or` chain.

**Documentation**
- The README subsection is concise and explains the practical implication (one-shot overrides work, `source .lnb.env` for normal project use). Nice touch.

**Minor observation (non-blocking)**
- The resolution logic now lives in two places (`get_notebook_dir` and the best-effort block in `main`). Not a problem today, but if a third caller ever appears, consider extracting a `_resolve_notebook_dir() -> Path | None` helper so the precedence rule has one source of truth. Flagging, not requesting.

[1 inline comment left.]

PR #19, opened 2026-05-10
feat: incremental ingest via per-file byte offsets

carbonscott:
# Incremental rebuild — implementation handoff

## Goal

Replace today's "full DELETE + re-parse every JSONL on every stale read" with per-file byte-offset incremental ingest. Full rebuild stays as the always-correct fallback whenever an invariant is broken. Why this framing: O(N) re-parse on every read is the only architectural cost in the current "JSONL is truth, SQLite is disposable" design (cli.py:343-367); the ceiling moves from "tens of thousands of entries" to "as much as you want" with a small, contained change.

## Current state

Design is settled and illustrated at `handoff/incremental-rebuild.html` (open in a browser; side-by-side before/after, fences, EOF safety rule). No code changes yet. The change is additive — existing `_atomic_rebuild` / `rebuild_from_jsonl` / `_index_is_stale` are preserved as the fallback path. The work is: (1) add the `_ingest_state` table DDL, (2) write a new `incremental_ingest` function alongside `rebuild_from_jsonl`, (3) branch in `ensure_db`, (4) tests.

## Decisions (with why)

- **Add `_ingest_state(file TEXT PRIMARY KEY, offset INTEGER NOT NULL)` to `index.sqlite`.** Append its `CREATE TABLE IF NOT EXISTS` to the existing `build_sql.create_sql` string (cli.py:146-156). State lives with the index; one artifact = impossible to drift; `_atomic_rebuild`'s tempfile-and-rename naturally clears it on full rebuild.

- **Reuse the existing `upsert_entry` (cli.py:283-297) from the new incremental path; do not write a parallel insert.** It already handles re-insert idempotently: `INSERT OR REPLACE` on `entries`, plus `DELETE FROM entries_fts WHERE rowid = ?` then `INSERT` for FTS5. So FTS5 dedupe is already solved by the existing pattern — no `INSERT OR IGNORE`, no rowid juggling needed in new code.

- **Per-file transaction shape: `BEGIN; upsert each new line; UPDATE _ingest_state SET offset = last_safe_pos; COMMIT;`.** Offset only advances on successful commit. Rollback leaves a clean prior state; next read retries from the prior offset. Idempotent retries are safe via the previous decision.

- **Advance offset only to the byte position after the last successfully-parsed, `\n`-terminated line in the new region.** A partial trailing line — from a crashed emit or an in-flight write during read — must NOT be skipped past. Leaving the offset before it means the next read retries that region; the writer's eventual completion is then ingested. This is the only correctness landmine in the design.

- **No special-case migration branch for existing `index.sqlite` files.** On first run with the new code, `executescript(sql.create)` creates `_ingest_state` empty. Absent rows = offset 0. The incremental path naturally reads each JSONL from byte 0 the first time (equivalent work to one full re-ingest), then incremental thereafter. No version detection, no migration code.

- **`cmd_rebuild` is unchanged (cli.py:628-634).** It calls `_atomic_rebuild`, which writes a fresh `index.sqlite` via tempfile + atomic rename. The fresh index has an empty `_ingest_state` because the modified `create_sql` includes the new DDL. Manual escape hatch stays as-is.

- **No new user-facing flag** (no `--incremental`, no `--no-incremental`). The system is always correct; incremental is just sometimes faster. A flag means documenting and supporting it forever for an optimization users shouldn't think about.

- **Schema change → full rebuild, not incremental ALTER.** ALTER + FTS5 virtual-table surgery is a tar pit for marginal gain. The existing `schema.yaml` mtime ≥ index mtime check in `_index_is_stale` (cli.py:300-311) is the trigger; route it into `_atomic_rebuild` as today.

- **Strict scope: do NOT modify `cmd_search` (cli.py:595-617).** A two-stage `bm25()`→`ORDER BY ts DESC` rerank is a separate, parked change. Bundling it would obscure the diff and complicate review.

## Ruled out (with why)

- **Sidecar JSON file for offsets** (e.g. `.lnb/_ingest_state.json`). Two artifacts that must stay in sync; either can be deleted independently and drift. SQLite table is atomic with the index it describes.

- **Line numbers instead of byte offsets.** Counting lines requires reading from byte 0 every time. Defeats the entire optimization.

- **`mtime` column in `_ingest_state`.** Redundant with the `size < offset` fence — adds a column and offers nothing extra.

- **Per-`emit` SQLite write** (emit also inserts into the index live, alongside the JSONL append). Violates "JSONL is truth, SQLite is disposable" — adds write-side coordination across multiple processes; deviates from the convergent agent-memory pattern documented in `handoff/jsonl-memory-research.md`.

## Load-bearing assumptions

- **JSONL files are append-only in normal use.** Hand-edits, truncation, external rewrites are exceptional and trip the `size < offset` fence into full rebuild. Don't try to reconcile non-append edits.

- **`entries.id` is globally unique** across the notebook (generated as `YYYYMMDDTHHMMSS-<4-hex>` by `generate_id`). The existing `INSERT OR REPLACE` pattern's correctness depends on this.

- **Single writer per JSONL file** (per-writer file model — see cli.py:573, `f"{writer_id}.jsonl"`). Different writers go to different files. No inter-file race exists or needs to be invented.

- **Concurrent readers may both run incremental ingest on the same index.** SQLite serializes via its own locking; do not add file locks or OS-level mutexes. Worst case: same lines `INSERT OR REPLACE`d twice. Wasteful but correct.

- **The schema-change fence uses `schema.yaml` mtime ≥ index mtime** as today (cli.py:300-311). Continue using it; do not invent a new check.

- **This is an addition alongside existing functions, not a rewrite.** `_atomic_rebuild`, `rebuild_from_jsonl`, `_index_is_stale` stay as the fallback. The new function is a separate `incremental_ingest` (or similarly named) called from `ensure_db` when the schema fence passes.

## Open threads

- **Orphan `_ingest_state` rows** (a JSONL file was deleted from disk but its row remains). Harmless — never matches a live file — but a one-line cleanup (`DELETE FROM _ingest_state WHERE file NOT IN (...)`) could run at the start of incremental ingest. Not blocking; intern judgment call.

- **Whether `format_schema_help()` (cli.py:58-80) should mention `_ingest_state`.** Probably no — it's internal — but worth a moment of thought during implementation. Easy to revisit if a user complaint surfaces.

- **Two-stage `search` rerank** (BM25 candidates → `ORDER BY ts DESC` via a CTE around `bm25(entries_fts)`). Adjacent to this work in the codebase, confirmed as the right shape, parked as a separate change. Out of scope here; OK to flag for follow-up in the PR description.

- **Minimal `parent_id` / graph-shape addition** (a single optional schema field for causal entry chains). Discussed earlier, parked. Out of scope.

## Pointers

- **`handoff/incremental-rebuild.html`** — primary visual reference. Side-by-side before/after, fences, EOF safety. Read first.
- **`handoff/jsonl-memory-research.md`** — motivates "JSONL is truth, SQLite is disposable." Read for context on *why* the rebuild is read-triggered, not write-triggered.
- **`src/lab_notebook/cli.py:125-178`** (`build_sql`) — where the new `_ingest_state` `CREATE TABLE IF NOT EXISTS` gets appended to `create_sql`. The `executescript` consumer at `_atomic_rebuild` (cli.py:321) handles multi-statement strings already.
- **`src/lab_notebook/cli.py:283-297`** (`upsert_entry`) — existing idempotent row insert + FTS5 re-insert. Call this from the new incremental path with `commit=False`; do one commit per file together with the offset update.
- **`src/lab_notebook/cli.py:300-340`** (`_index_is_stale`, `_atomic_rebuild`, `ensure_db`) — the integration point. The new branching lives in `ensure_db`: if the index exists and the schema fence passes, call the new incremental function; otherwise fall back to `_atomic_rebuild` exactly as today.
- **`src/lab_notebook/cli.py:343-367`** (`rebuild_from_jsonl`) — the existing loop the new `incremental_ingest` mirrors (per-file `for line in f:` + `upsert_entry`), but with `seek(off)` and offset bookkeeping.
- **`src/lab_notebook/cli.py:524-579`** (`cmd_emit`) — confirms the emit invariant: JSONL append + `fsync` only, no SQLite write. The new ingest must not require any change to emit.
- **`tests/test_cli.py`** — existing test patterns; new tests belong here. Coverage to aim for: fresh notebook → first read; emit-then-read incremental fast path; partial-line-at-EOF (don't advance past it; complete the line; next read picks it up); JSONL truncated → full rebuild fallback; `schema.yaml` newer than index → full rebuild fallback; existing pre-`_ingest_state` index opened by new code → re-ingests from offset 0 with no special-case code.

### DIFF
--- a/src/lab_notebook/cli.py
+++ b/src/lab_notebook/cli.py
@@ -299,20 +299,6 @@ def upsert_entry(conn: sqlite3.Connection, entry: dict, sql: SchemaSQL,
         conn.commit()
 
 
-def _index_is_stale(notebook_dir: Path, dbp: Path) -> bool:
-    """True if any JSONL file or schema.yaml is newer than the index."""
-    idx_mtime = dbp.stat().st_mtime
-    schema_file = notebook_dir / "schema.yaml"
-    if schema_file.exists() and schema_file.stat().st_mtime >= idx_mtime:
-        return True
-    edir = entries_dir(notebook_dir)
-    if edir.exists():
-        for f in edir.glob("*.jsonl"):
-            if f.stat().st_mtime >= idx_mtime:
-                return True
-    return False
-
-
 def _atomic_rebuild(notebook_dir: Path, dbp: Path,
                      schema: dict, sql: SchemaSQL) -> int:
     """Rebuild the index into a temp file, then atomically rename into place."""
@@ -418,18 +404,21 @@ def ensure_db(notebook_dir: Path, schema: dict | None = None) -> tuple[sqlite3.C
         print(f"Index rebuilt: {count} entries", file=sys.stderr)
     else:
         conn = sqlite3.connect(str(dbp))
+        fence_tripped = None
+        new_count = 0
         try:
             new_count = incremental_ingest(conn, notebook_dir, schema, sql)
         except _IngestFenceTripped as e:
+            fence_tripped = str(e)
+        finally:
             conn.close()
-            print(f"Ingest fence tripped on {e}; falling back to full rebuild",
-                  file=sys.stderr)
+        if fence_tripped is not None:
+            print(f"Ingest fence tripped on {fence_tripped}; "
+                  f"falling back to full rebuild", file=sys.stderr)
             count = _atomic_rebuild(notebook_dir, dbp, schema, sql)
             print(f"Index rebuilt: {count} entries", file=sys.stderr)
-        else:
-            conn.close()
-            if new_count:
-                print(f"Index updated: +{new_count} entries", file=sys.stderr)
+        elif new_count:
+            print(f"Index updated: +{new_count} entries", file=sys.stderr)
     conn = sqlite3.connect(str(dbp))
     return conn, schema, sql
 
--- a/src/lab_notebook/cli.py
+++ b/src/lab_notebook/cli.py
@@ -153,6 +153,8 @@ def build_sql(schema: dict) -> SchemaSQL:
         + "CREATE INDEX IF NOT EXISTS idx_entries_context ON entries(context);\n"
         + "CREATE INDEX IF NOT EXISTS idx_entries_type ON entries(type);\n"
         + "CREATE INDEX IF NOT EXISTS idx_entries_ts ON entries(ts);\n"
+        + "CREATE TABLE IF NOT EXISTS _ingest_state ("
+        + "file TEXT PRIMARY KEY, offset INTEGER NOT NULL);\n"
     )
 
     # -- UPSERT --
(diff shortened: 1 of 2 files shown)


src/lab_notebook/cli.py
```diff
@@ -328,43 +330,126 @@ def _atomic_rebuild(notebook_dir: Path, dbp: Path,
...
     if schema is None:
         schema = load_schema(notebook_dir)
     sql = build_sql(schema)
     dbp = index_path(notebook_dir)
-    if not dbp.exists() or _index_is_stale(notebook_dir, dbp):
+    if not dbp.exists() or _schema_newer_than_index(notebook_dir, dbp):
         count = _atomic_rebuild(notebook_dir, dbp, schema, sql)
         print(f"Index rebuilt: {count} entries", file=sys.stderr)
+    else:
+        conn = sqlite3.connect(str(dbp))
```
carbonscott, line 406: Optional: If `incremental_ingest` raises anything other than `_IngestFenceTripped` (e.g. `sqlite3.OperationalError`, an `OSError` from `stat`/`open`, etc.), this `conn` is leaked because only the fence path and the `else` branch close it. The exception then propagates out of `ensure_db` with the connection still open. A `try/finally` (or `with contextlib.closing(conn):`) would make the close path unconditional. Minor — `incremental_ingest` already rolls back its per-file transaction, so durability is fine; just the file handle leaks until GC.

src/lab_notebook/cli.py
```diff
@@ -328,43 +330,126 @@ def _atomic_rebuild(notebook_dir: Path, dbp: Path,
...
+                    print(
+                        f"Warning: {jsonl_file.name}: skipping malformed line: {e}",
+                        file=sys.stderr,
+                    )
+                    last_safe_pos = line_end_pos
+                    continue
+                flat = flatten_entry(entry, schema)
+                upsert_entry(conn, flat, sql, commit=False)
+                total += 1
+                last_safe_pos = line_end_pos
```
carbonscott, line 385: FYI: Setting `last_safe_pos = line_end_pos` only AFTER the successful `upsert_entry` (and likewise after the blank/malformed branches that explicitly chose to skip) is the correctness pivot for this whole feature. A crash between `upsert_entry` and the offset commit is safe because the next read re-processes the same region and `INSERT OR REPLACE` makes that idempotent. Worth keeping that invariant in mind for any future change to this loop — easy to accidentally hoist this assignment above the parse and silently introduce a skip-past-broken-line bug.

tests/test_cli.py
```diff
@@ -596,6 +596,180 @@ def test_stale_after_schema_change(self, notebook):
...
+    def test_pre_ingest_state_index_migrates_transparently(self, notebook):
+        cmd_emit(make_emit_args(content="entry one"))
+        cmd_emit(make_emit_args(content="entry two"))
+        ensure_db(notebook)[0].close()
+        # Simulate an existing pre-_ingest_state index by dropping the table.
+        conn = sqlite3.connect(str(index_path(notebook)))
+        conn.execute("DROP TABLE _ingest_state")
+        conn.commit()
+        conn.close()
+        # Touch the JSONL so the schema fence is not the trigger here — but the
```
carbonscott: Nit: This comment talks about "touching the JSONL so the schema fence is not the trigger here" but the test never actually touches the JSONL or the schema. Looks like leftover prose from an earlier draft. The test itself is correct (drop `_ingest_state`, run `ensure_db`, verify transparent re-ingest); just the comment can be trimmed.

carbonscott (commented):
## Review Summary

**Overall:** Solid, focused incremental-rebuild change. The design is clean (single artifact, no migration code, idempotent re-ingest), the EOF-safety logic is carefully thought through, and test coverage is thorough across the failure modes. Approving.

### Key Findings

**Design** — The decisions documented in the PR description are well reasoned. Co-locating `_ingest_state` with the index, reusing the existing `upsert_entry` for idempotency, and keeping `_atomic_rebuild` as the always-correct fallback are all the right calls. The choice to not add a user-facing flag is correct — incremental should be invisible to users.

**Functionality** — The EOF-safety rule (advance offset only past complete `\n`-terminated lines that were either ingested, blank, or skipped-as-malformed) is implemented correctly. Concurrent-reader correctness is preserved via SQLite locking + `INSERT OR REPLACE` idempotency, as the description notes.

**Dead code** — `_index_is_stale` (`src/lab_notebook/cli.py:302-313`) is no longer called from production. Its job is now split: the schema fence moved to `_schema_newer_than_index`, and the JSONL-mtime branch is obsoleted by the no-op-when-`size == offset` fast path inside `incremental_ingest`. The function is still imported and asserted by tests at `tests/test_cli.py:577,587,596`. Worth either removing it (and updating those three tests) or leaving a one-liner explaining why it stays. Not blocking — flagging because dead code accumulates.

**Tests** — Coverage is excellent: first-read offset record, fast path, partial-line-at-EOF (both halves), truncation fence, schema fence, pre-`_ingest_state` migration, malformed-line-in-middle, per-writer offset isolation. Test names communicate intent well. `capsys` assertions on the stderr messages are a nice way to verify which code path actually ran.

**Stderr noise** — `Index updated: +N entries` is gated on `if new_count:`, so reads with no new content stay silent. Good.

### PR Description Feedback
Description is excellent — clearly states the goal, the why, the decisions taken, what was ruled out and why, load-bearing assumptions, open threads, and pointers into the code. This is a model PR description.

[3 inline comments left.]

src/lab_notebook/cli.py
carbonscott: Addressed in follow-up. Restructured `ensure_db` with `try/except/finally` — `finally: conn.close()` runs on every path, the fence trigger is captured into a local (`fence_tripped`), and the rebuild is dispatched **after** `conn.close()` so the previous "close before `_atomic_rebuild`" ordering is preserved (defensive against `os.rename` against an open handle on non-Linux platforms). Non-fence exceptions (`sqlite3.OperationalError`, `OSError` from `stat`/`open`) now propagate with the connection guaranteed closed.

src/lab_notebook/cli.py
carbonscott: Agreed — that ordering is the correctness pivot, and the only-after-success placement is intentional. No change.

tests/test_cli.py
carbonscott: Good catch — trimmed to a single one-liner pointing at the idempotent `INSERT OR REPLACE` behavior in the follow-up.

carbonscott:
Thanks for the review! Pushed a follow-up (294f565) addressing the actionable items:

- **Connection leak in `ensure_db`** — restructured with `try/except/finally` so the SQLite handle is closed on every path; fence trigger captured into a local and rebuild dispatched after `conn.close()` (preserves the previous "close before `_atomic_rebuild`" ordering). Non-fence exceptions (`sqlite3.OperationalError`, `OSError` from `stat`/`open`) can no longer leak the handle. Replied inline.
- **Dead `_index_is_stale`** (Key Finding) — removed, along with the `TestStaleness` class (3 tests) and the import. The behaviors those tests covered are now exercised by `TestIncrementalIngest::test_schema_change_falls_back_to_full_rebuild` (schema fence) and the offset-based tests (the `size == offset` fast path replaces the JSONL-mtime check).
- **Stale comment at `tests/test_cli.py:716`** — trimmed to a one-liner. Replied inline.

The `last_safe_pos` FYI was left as-is (acknowledged inline — the only-after-success placement is intentional).

`uv run pytest tests/test_cli.py -v` — 95 passed (was 98 − 3 deleted staleness tests).

carbonscott (commented):
## Follow-up Review

**Overall:** All three points from the previous review are addressed cleanly in `294f565`. Approving.

### Confirmed fixes

- **Dead code removed.** `_index_is_stale` is gone from `cli.py`; the import and `TestStaleness` class are gone from `tests/test_cli.py`. The schema-fence behavior is still exercised by `test_schema_change_falls_back_to_full_rebuild`, and the no-op fast path inside `incremental_ingest` covers the rest.

- **`ensure_db` close path.** The `try/except/finally` restructure is actually cleaner than the `try/finally` I suggested: capturing the fence trigger into a local (`fence_tripped`) lets the `finally` close the connection unconditionally, and the rebuild is dispatched after the close — preserving the original "close before `_atomic_rebuild`" ordering. Non-fence exceptions (e.g. `sqlite3.OperationalError`, `OSError` from `stat`/`open`) now propagate with the connection guaranteed closed.

- **Stale test comment trimmed.** The replacement (`Re-open: incremental_ingest recreates the table and re-ingests via idempotent INSERT OR REPLACE.`) is short and accurate.

Ready to merge from my side.

PR #20, opened 2026-06-01
feat: retract entries via append-only tombstones

carbonscott:
## What

Adds a `retract ID --reason "why"` command so the notebook can be used as *system memory* — wrong or out-of-date entries can be forgotten — without breaking the append-only model.

```
lab-notebook retract 20260321T143022-a7f2 --reason "superseded by corrected measurement"
```

## How it works

- **Append-only logical delete.** Retract appends a `_retract` tombstone record (`{type, retracts, reason}`) to the retracting writer's JSONL. On the next ingest the target row + its FTS row are hard-deleted from `index.sqlite`. The original entry and the tombstone both stay in the JSONL as the audit trail of what was forgotten and why.
- **Survives rebuild.** Because the index is derived from the JSONL, the deletion is reapplied on every rebuild.
- **Order-independent.** `incremental_ingest` is now a two-phase, single-commit pass: scan all files collecting tombstone targets + offsets, then apply deletes and offset writes in one transaction. This handles the case where a tombstone in one writer's file targets an entry in a later-sorted writer's file (any writer may retract any id), which on a full rebuild is only inserted later in the same pass. The single commit keeps a tombstone's offset advance atomic with its delete, so a crash can't strand a target.

## Design decisions (confirmed during planning)

- **Retract only** — no supersede/amend verb; corrections = retract + emit-new.
- **Hard delete from the index** — no soft status / `--show-retracted`. Recover by restoring from a JSONL backup and rebuilding.
- **Reason required** on every retract.
- **Anyone retracts anything** — no ownership check.

## Also

`_atomic_rebuild` now reports the **net active row count** (`COUNT(*)` after ingest) rather than lines ingested, so `Rebuilt index: N entries` reflects what's actually queryable after tombstones are applied.

## Tests

10 new tests (`TestRetract`): removal from search/FTS, tombstone-not-queryable, tombstone-preserved-in-JSONL, reason-required, nonexistent/already-retracted errors, cross-writer-survives-rebuild (the ordering guarantee), incremental-path offset advance, and net-count reporting. Full suite: 105 passing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,21 @@ file on disk (mutually exclusive with `--template`).
 Write a notebook entry. Required: `--context`, `--type`, content.
 Custom field flags (e.g. `--repo`, `--tags`) are generated from `schema.yaml`.
 
+### `retract ID --reason "why"`
+
+Forget an entry that is wrong or out of date. Both `ID` and `--reason` are
+required. Retract appends a tombstone record to the notebook — it does **not**
+edit or remove the original line. On the next indexed read the target row is
+deleted from `index.sqlite` (and from full-text search); the tombstone itself
+is never returned as an entry.
+
+Because the index is rebuilt from the JSONL, the deletion survives `rebuild`.
+The retraction is logical, not physical: the original entry and the tombstone
+(with its reason) both remain in `entries/*.jsonl` as the audit trail of what
+was forgotten and why. To recover a retracted entry, restore it from a JSONL
+backup and `rebuild` — there is no un-retract command. Any writer may retract
+any entry by id.
+
 ### `sql "query"`
 
 Run raw SQL against the index. Auto-rebuilds if `index.sqlite` is missing.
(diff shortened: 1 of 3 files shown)


carbonscott (commented):
## Review Summary

**Overall:** Approve. Well-designed, well-tested addition of logical delete to the append-only model. The previously-reported tuple-mismatch regression and its siblings are fully fixed, with a targeted regression test, and the suite passes (109 tests).

### Verified
- Two-phase, single-commit incremental ingest is correct: order-independent across writer files, and each tombstone's offset advance is atomic with its delete.
- Both early-exit paths return tuples (`incremental_ingest` and `rebuild_from_jsonl` -> `return 0, 0`); every consumer checked, no scalar-unpack mismatch remains.
- Regression test `test_incremental_ingest_missing_entries_dir` covers the exact "index exists but entries/ absent" path that previously raised TypeError.
- `delete_entry` mirrors the established FTS-by-rowid bookkeeping in `upsert_entry`.
- Reserved-type guard rejects `_retract` in schema.yaml, preventing silently-swallowed rows.

### Non-blocking (FYI)
- The deferred-offset safety invariant holds for CLI-written tombstones; a hand-written tombstone referencing a not-yet-ingested id is outside the supported workflow and is reconciled by a full rebuild.
- PR description test count ("10 new tests / 105 passing") is slightly stale vs. the current 109.

No blocking issues.

PR #31, opened 2026-07-04
feat: shell tab-completion for the CLI (bash)

carbonscott:
## Summary

Adds bash tab-completion for the `lab-notebook` CLI, surfacing schema-constrained
values (types, contexts, field names, prior field values) at Tab-time so users
don't have to recall them by hand. Zero new runtime deps.

**Install:**
```bash
source <(lab-notebook completion bash)
```

**What completes**

| At | Completes to | Source |
|----|--------------|--------|
| `lab-notebook <TAB>` | subcommands (+`completion`) | static |
| `emit --type <TAB>` | entry types | `schema.yaml` (no DB) |
| `emit/search --context <TAB>` | distinct contexts | read-only index |
| `emit -f <TAB>` | field names as `name=` | `schema.yaml` |
| `emit -f repo=<TAB>` | distinct existing values | read-only index |
| `<subcmd> -<TAB>` | that subcommand's flags | static |

## Design

- A hidden `__complete` subcommand prints newline-delimited candidates; a thin
  bash function (`completion bash`) feeds them to `compgen` at Tab-time. All
  knowledge lives in Python (`complete.py`); the shell stays dumb.
- **Strictly read-only.** New `store.open_readonly` opens the index in `mode=ro`
  and never rebuilds or ingests — pressing Tab has no side effects and never
  prints an "Index rebuilt" line. Missing/unresolved notebook or missing index
  → the affected slot offers nothing (schema-only slots still work without a DB).
- The `=` bash wordbreak is handled so `-f repo=<value>` completes the value
  while leaving `repo=` attached (via `compopt -o nospace` on `name=` candidates).
- Bash only for v1; the Python backend is shared, so zsh is a cheap follow-up.
- Known limitation (documented): field values containing spaces/glob metacharacters
  don't complete cleanly. Contexts/types/keys/tags are space-free in practice.

## Files

- `src/lab_notebook/complete.py` — **new**: slot logic + live resolvers.
- `src/lab_notebook/store.py` — `open_readonly` (first read-only opener).
- `src/lab_notebook/cli.py` — `completion` + hidden `__complete` subparsers/handlers, `BASH_COMPLETION_SCRIPT`.
- `tests/test_cli.py` — `TestCompletion` (16 tests).
- `README.md`, `skill/SKILL.md` — document the feature (same-PR docs convention).

## Testing

- `uv run pytest -q` → **155 passed** (139 prior + 16 new).
- New tests cover every slot, graceful degradation (unresolved dir / missing
  index → `[]`, no raise), and a no-rebuild proof (index mtime byte-for-byte
  unchanged, no "Index rebuilt" output after exercising every path).
- Verified end-to-end through the actual `complete -F`/`compgen` bash function,
  including the `=` wordbreak remap and `nospace`, and confirmed via
  `stat` that a Tab session leaves `index.sqlite` untouched.

https://claude.ai/code/session_01DJMM712b8HgapincVAheha

### DIFF
--- a/src/lab_notebook/cli.py
+++ b/src/lab_notebook/cli.py
@@ -11,6 +11,7 @@ Usage:
     lab-notebook rebuild
     lab-notebook contexts
     lab-notebook template [name] [--force]
+    lab-notebook completion bash
 """
 from __future__ import annotations
 
@@ -40,6 +41,7 @@ from .store import (
     get_notebook_dir,
     index_path,
 )
+from .complete import complete
 
 EXAMPLE_QUERIES = """\
 Example queries
@@ -59,6 +61,33 @@ WHERE entries_fts MATCH 'search term';
 -- Entries per context
 SELECT context, COUNT(*) AS n, MIN(ts) AS first, MAX(ts) AS latest
 FROM entries GROUP BY context ORDER BY latest DESC;
+"""
+
+# Bash registration script printed by `lab-notebook completion bash`. Install
+# with `source <(lab-notebook completion bash)`. The function shells out to the
+# hidden `__complete` subcommand at Tab-time; all completion knowledge lives in
+# Python (see complete.py), the shell stays dumb.
+#
+# '=' is in bash's default COMP_WORDBREAKS, so `-f repo=ma` tokenizes to
+# `… -f repo = ma`: Python keys off the '=' token to find the field, and bash
+# completes only the post-'=' word (`ma`→`mae`), leaving `repo=` intact. The
+# `cur=="="` guard covers the empty-value case (`-f repo=<TAB>`). Passing `--`
+# before the words makes argparse treat flag-like tokens (`--type`, `-f`) as
+# positionals rather than options.
+BASH_COMPLETION_SCRIPT = r"""_lab_notebook_complete() {
+    local cur="${COMP_WORDS[COMP_CWORD]}"
+    [[ "$cur" == "=" ]] && cur=""          # '=' wordbreak: empty value after key=
+    local IFS=$'\n'
+    local candidates
+    candidates="$(lab-notebook __complete "$COMP_CWORD" -- "${COMP_WORDS[@]}" 2>/dev/null)"
+    COMPREPLY=( $(compgen -W "$candidates" -- "$cur") )
+    # Field-key candidates end with '=' — keep the cursor attached so the value
+    # can be typed immediately. Only suppress the trailing space in that case.
+    if [[ ${#COMPREPLY[@]} -gt 0 && "${COMPREPLY[0]}" == *= ]]; then
+        compopt -o nospace
+    fi
+}
+complete -F _lab_notebook_complete lab-notebook
 """
 
 
@@ -301,6 +330,23 @@ def cmd_template(args: argparse.Namespace) -> None:
         print("Run 'lab-notebook rebuild' to re-index existing entries.")
 
 
+def cmd_completion(args: argparse.Namespace) -> None:
+    # args.shell is constrained to "bash" by argparse choices.
+    print(BASH_COMPLETION_SCRIPT, end="")
+
+
+def cmd___complete(args: argparse.Namespace) -> None:
+    # Internal RPC target for the shell function. Must never raise on odd input
+    # (a Tab press should never spew a traceback), so swallow everything and
+    # emit nothing on failure. complete() already returns [] for the expected
+    # degradation paths; this also guards genuinely malformed argv.
+    try:
+        for candidate in complete(args.words, args.cword):
+            print(candidate)
+    except Exception:
+        pass
+
+
 # ---------------------------------------------------------------------------
 # CLI
 # ---------------------------------------------------------------------------
@@ -387,6 +433,18 @@ def main() -> None:
     p_template.add_argument("--force", action="store_true",
                             help="Overwrite existing schema.yaml")
     p_template.set_defaults(func=cmd_template)
+
+    # -- completion -- (prints a shell registration script)
+    p_completion = sub.add_parser("completion", help="Print a shell completion script")
+    p_completion.add_argument("shell", choices=["bash"])
+    p_completion.set_defaults(func=cmd_completion)
+
+    # -- __complete -- (hidden RPC target for the shell function; no help= so it
+    # stays out of the visible subcommand listing)
+    p_complete = sub.add_parser("__complete")
+    p_complete.add_argument("cword", type=int)
+    p_complete.add_argument("words", nargs="*")
+    p_complete.set_defaults(func=cmd___complete)
 
     args = parser.parse_args()
     try:
--- a/src/lab_notebook/cli.py
+++ b/src/lab_notebook/cli.py
@@ -450,8 +450,10 @@ def build_parser() -> argparse.ArgumentParser:
     p_completion.add_argument("shell", choices=["bash"])
     p_completion.set_defaults(func=cmd_completion)
 
-    # -- __complete -- (hidden RPC target for the shell function; no help= so it
-    # stays out of the visible subcommand listing)
+    # -- __complete -- (internal RPC target for the shell function; no help= so
+    # it carries no description in --help. The name still appears in the usage
+    # metavar — argparse has no clean way to drop one choice from it — but users
+    # are not meant to invoke it directly.)
     p_complete = sub.add_parser("__complete")
     p_complete.add_argument("cword", type=int)
     p_complete.add_argument("words", nargs="*")
--- a/src/lab_notebook/cli.py
+++ b/src/lab_notebook/cli.py
@@ -68,19 +68,30 @@ FROM entries GROUP BY context ORDER BY latest DESC;
 # hidden `__complete` subcommand at Tab-time; all completion knowledge lives in
 # Python (see complete.py), the shell stays dumb.
 #
+# Candidates are read one-per-line straight into COMPREPLY and are NEVER routed
+# through `compgen -W`. `compgen -W` re-expands every word in its list — command
+# substitution, backticks, globbing — so a candidate drawn from untrusted
+# notebook data (a context, tag, or field value like `$(rm -rf ~)` or `*`) would
+# be executed or glob-expanded the instant a user pressed Tab. Instead, a
+# `while read` loop assigns each line literally and does the prefix match in the
+# shell with a quoted pattern (`"$cur"*`), so every candidate stays inert text.
+# `read` + process substitution also keeps this working on bash 3.2 (no
+# `mapfile`).
+#
 # '=' is in bash's default COMP_WORDBREAKS, so `-f repo=ma` tokenizes to
-# `… -f repo = ma`: Python keys off the '=' token to find the field, and bash
-# completes only the post-'=' word (`ma`→`mae`), leaving `repo=` intact. The
-# `cur=="="` guard covers the empty-value case (`-f repo=<TAB>`). Passing `--`
-# before the words makes argparse treat flag-like tokens (`--type`, `-f`) as
-# positionals rather than options.
+# `… -f repo = ma`: Python keys off the '=' token to find the field, and the
+# in-shell prefix match completes only the post-'=' word (`ma`→`mae`), leaving
+# `repo=` intact. The `cur=="="` remap covers the empty-value case
+# (`-f repo=<TAB>`). Passing `--` before the words makes argparse treat
+# flag-like tokens (`--type`, `-f`) as positionals rather than options.
 BASH_COMPLETION_SCRIPT = r"""_lab_notebook_complete() {
     local cur="${COMP_WORDS[COMP_CWORD]}"
     [[ "$cur" == "=" ]] && cur=""          # '=' wordbreak: empty value after key=
-    local IFS=$'\n'
-    local candidates
-    candidates="$(lab-notebook __complete "$COMP_CWORD" -- "${COMP_WORDS[@]}" 2>/dev/null)"
-    COMPREPLY=( $(compgen -W "$candidates" -- "$cur") )
+    local line
+    COMPREPLY=()
+    while IFS= read -r line; do
+        [[ "$line" == "$cur"* ]] && COMPREPLY+=("$line")
+    done < <(lab-notebook __complete "$COMP_CWORD" -- "${COMP_WORDS[@]}" 2>/dev/null)
     # Field-key candidates end with '=' — keep the cursor attached so the value
     # can be typed immediately. Only suppress the trailing space in that case.
     if [[ ${#COMPREPLY[@]} -gt 0 && "${COMPREPLY[0]}" == *= ]]; then
@@ -351,7 +362,7 @@ def cmd___complete(args: argparse.Namespace) -> None:
 # CLI
 # ---------------------------------------------------------------------------
 
-def main() -> None:
+def build_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(
         prog="lab-notebook",
         description="Lab notebook: append-only JSONL + SQLite query index",
@@ -446,6 +457,11 @@ def main() -> None:
     p_complete.add_argument("words", nargs="*")
     p_complete.set_defaults(func=cmd___complete)
 
+    return parser
+
+
+def main() -> None:
+    parser = build_parser()
     args = parser.parse_args()
     try:
         args.func(args)
--- a/src/lab_notebook/complete.py
+++ b/src/lab_notebook/complete.py
@@ -93,7 +93,10 @@ def _contexts():
             "WHERE context IS NOT NULL ORDER BY 1"
         ).fetchall()
         return [str(r[0]) for r in rows]
-    except sqlite3.OperationalError:
+    except sqlite3.Error:
+        # Any DB-layer error (missing table on a partial index, or a corrupt
+        # file raising the OperationalError parent DatabaseError) → offer
+        # nothing. A Tab press must never surface a traceback.
         return []
     finally:
         conn.close()
@@ -143,7 +146,9 @@ def _field_values(field):
             f"WHERE {col} IS NOT NULL ORDER BY 1"
         ).fetchall()
         return [str(r[0]) for r in rows]
-    except sqlite3.OperationalError:
+    except sqlite3.Error:
+        # See _contexts: swallow any DB-layer error (incl. the corrupt-file
+        # DatabaseError, which is OperationalError's parent) so Tab stays quiet.
         return []
     finally:
         conn.close()
@@ -157,32 +162,44 @@ def complete(words: list[str], cword: int) -> list[str]:
     """Return candidates for the word at index `cword`.
 
     `words[0]` is the program name (`lab-notebook`); `cword` indexes the word
-    being completed. Never raises: unresolvable notebook / missing schema /
-    missing index all yield [].
+    being completed. Never raises: any out-of-range `cword`, unresolvable
+    notebook, missing schema, or missing index all yield []. (The real bash
+    driver always passes an in-range COMP_CWORD, but this stays total so a
+    malformed `__complete` argv degrades to silence rather than a traceback.)
     """
     if cword <= 1:
         return SUBCOMMANDS
+    if len(words) < 2:
+        return []              # cword past the end with no subcommand typed yet
     sub = words[1]
-    cur = words[cword] if cword < len(words) else ""
-    prev = words[cword - 1] if cword >= 1 else ""
+    sub_flags = FLAGS.get(sub, [])
+    # Bounds-guard cur/prev independently — cword may point past the end (a
+    # trailing-space completion) or, defensively, anywhere.
+    cur = words[cword] if 0 <= cword < len(words) else ""
+    prev = words[cword - 1] if 0 <= cword - 1 < len(words) else ""
 
     # --- value slots first (before flag-name) ---
+    # Each value slot is gated on the triggering flag actually belonging to
+    # `sub`, matching the subcommand-scoped flag-name slot below — so a
+    # flag/subcommand mismatch (e.g. `sql --type <TAB>`) offers nothing.
+    #
     # `-f <field> = <cur>`: '=' is a bash COMP_WORDBREAK, so key/=/value are
     # separate words. Two shapes: prev=='=' (value being typed) or cur=='='
     # (empty value right after `key=`).
-    if prev == "=" and cword >= 3 and words[cword - 3] in ("-f", "--field"):
+    has_field = "-f" in sub_flags or "--field" in sub_flags
+    if has_field and prev == "=" and cword >= 3 and words[cword - 3] in ("-f", "--field"):
         return _field_values(words[cword - 2])
-    if cur == "=" and cword >= 2 and words[cword - 2] in ("-f", "--field"):
+    if has_field and cur == "=" and cword >= 2 and words[cword - 2] in ("-f", "--field"):
         return _field_values(words[cword - 1])
 
-    if prev == "--type":
+    if prev == "--type" and "--type" in sub_flags:
         return _types()
-    if prev == "--context":
+    if prev == "--context" and "--context" in sub_flags:
         return _contexts()
-    if prev in ("-f", "--field"):
+    if prev in ("-f", "--field") and has_field:
         return _field_keys()   # each ends with "="
 
     # --- flag-name slot ---
     if cur.startswith("-"):
-        return FLAGS.get(sub, [])
+        return sub_flags
     return []
--- a/src/lab_notebook/complete.py
+++ b/src/lab_notebook/complete.py
@@ -1,0 +1,188 @@
+"""Shell tab-completion backend for the `lab-notebook` CLI.
+
+A hidden `__complete` subcommand (see cli.py) prints newline-delimited
+candidates that a thin bash script feeds to `compgen` at Tab-time. All the
+knowledge lives here; the shell stays dumb.
+
+Direction stays one-way: schema <- store <- complete <- cli. This module
+depends only on `schema` (for `load_schema`) and `store` (for
+`get_notebook_dir` / `open_readonly`).
+
+Everything resolves LIVE from the target notebook (`schema.yaml` +
+`index.sqlite`), never from the bundled templates. All reads are strictly
+read-only: `open_readonly` never rebuilds or ingests, so tab-completion has no
+side effects and tolerates a stale index. Every resolver swallows `LnbError`
+and SQLite errors, so completion emits nothing rather than raising when the
+notebook is unresolved, the schema is missing, or the index does not exist yet.
+"""
+from __future__ import annotations
+
+import json
+import sqlite3
+
+from .schema import LnbError, load_schema
+from .store import get_notebook_dir, open_readonly
+
+# The subcommands a user can complete: the ten registered names plus
+# `completion` (the install helper). `__complete` is deliberately excluded — it
+# is an internal RPC target, not something a user types.
+SUBCOMMANDS = [
+    "init", "emit", "retract", "show", "sql", "search",
+    "schema", "rebuild", "contexts", "template", "completion",
+]
+
+# Per-subcommand flag names, offered in the flag-name slot (`<subcmd> -<TAB>`).
+FLAGS = {
+    "init": ["--template", "--template-path", "--force"],
+    "emit": ["--context", "--type", "--artifacts", "-f", "--field", "--extra"],
+    "retract": ["--reason"],
+    "search": ["--context", "--type"],
+    "template": ["--force"],
+}
+
+
+# ---------------------------------------------------------------------------
+# Live resolvers (all side-effect-free; degrade to []/None on any error)
+# ---------------------------------------------------------------------------
+
+def _dir():
+    """Resolve the target notebook dir, or None if it cannot be resolved."""
+    try:
+        return get_notebook_dir()
+    except LnbError:
+        return None
+
+
+def _schema():
+    """Load the target schema, or None if the dir/schema cannot be resolved."""
+    d = _dir()
+    if d is None:
+        return None
+    try:
+        return load_schema(d)
+    except LnbError:
+        return None
+
+
+def _types():
+    schema = _schema()
+    if schema is None:
+        return []
+    return [str(t) for t in schema["types"]]
+
+
+def _field_keys():
+    # Each key ends with "=" so the value can be typed immediately after Tab.
+    # `schema["fields"]` already includes the built-in `artifacts`.
+    schema = _schema()
+    if schema is None:
+        return []
+    return [f"{k}=" for k in schema["fields"]]
+
+
+def _contexts():
+    d = _dir()
+    if d is None:
+        return []
+    conn = open_readonly(d)
+    if conn is None:
+        return []
+    try:
+        rows = conn.execute(
+            "SELECT DISTINCT context FROM entries "
+            "WHERE context IS NOT NULL ORDER BY 1"
+        ).fetchall()
+        return [str(r[0]) for r in rows]
+    except sqlite3.OperationalError:
+        return []
+    finally:
+        conn.close()
+
+
+def _field_values(field):
+    """Distinct existing values for `field` from the index.
+
+    `field` must be a declared schema field — the check guards the double-quoted
+    column interpolation below (the same interpolation build_sql uses to create
+    the column, so a name that made it into the table is safe here). For a
+    `list` field, values are stored as JSON arrays; we json.loads and flatten in
+    Python (no json_each dependency). Any other type reads distinct raw values.
+    """
+    d = _dir()
+    if d is None:
+        return []
+    try:
+        schema = load_schema(d)
+    except LnbError:
+        return []
+    fields = schema["fields"]
+    if field not in fields:
+        return []
+    conn = open_readonly(d)
+    if conn is None:
+        return []
+    try:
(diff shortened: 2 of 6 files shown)


tests/test_cli.py
```diff
@@ -1811,3 +1813,183 @@ def test_retract_missing_target_raises(self, notebook):
...
+        complete(["lab-notebook", "search", "--context"], 3)
+        complete(["lab-notebook", "emit", "-f"], 3)
+        complete(["lab-notebook", "emit", "-f", "repo", "="], 4)
+        complete(["lab-notebook", "emit", "-f", "tags", "="], 4)
+        complete(["lab-notebook", "emit", "-f", "repo", "=", "ot"], 5)
+        complete(["lab-notebook", "emit", "-"], 2)
+        captured = capsys.readouterr()
+        assert "Index rebuilt" not in captured.err
+        assert "Index rebuilt" not in captured.out
+        assert dbp.stat().st_mtime_ns == mtime_before   # byte-for-byte untouched
```
carbonscott, line 1979: This is the crux of the "no-rebuild proof," but it runs against a *fresh* index built by `_seed_completion`, and on a fresh index the assertion is non-discriminating. I verified that `ensure_db()` (the normal, rebuild-capable read path) also leaves a fresh index's mtime unchanged and prints nothing -- so this test would still pass even if the resolvers were re-routed from `open_readonly` to `ensure_db`. The two paths only diverge on a *stale* index (schema.yaml newer than index.sqlite): there `open_readonly` reads untouched while `ensure_db` rebuilds (mtime changes + "Index rebuilt" on stderr). Since complete.py:14 and README:229 explicitly promise "tolerates a stale index," please add a stale-index variant (bump schema.yaml mtime past the index, then exercise the index-backed slots and assert mtime unchanged + no "Index rebuilt"). That gives the guarantee teeth against a regression to the rebuild-capable path.

src/lab_notebook/complete.py
```diff
@@ -0,0 +1,188 @@
...
+knowledge lives here; the shell stays dumb.
+
+Direction stays one-way: schema <- store <- complete <- cli. This module
+depends only on `schema` (for `load_schema`) and `store` (for
+`get_notebook_dir` / `open_readonly`).
+
+Everything resolves LIVE from the target notebook (`schema.yaml` +
+`index.sqlite`), never from the bundled templates. All reads are strictly
+read-only: `open_readonly` never rebuilds or ingests, so tab-completion has no
+side effects and tolerates a stale index. Every resolver swallows `LnbError`
```
carbonscott, line 14: "tolerates a stale index" is the one read-only behavior with no test. It is genuinely true in the code (`open_readonly` never checks staleness and never routes to `ensure_db`), but the only no-rebuild test uses a fresh index where the rebuild-capable path is indistinguishable. A stale-index test (see comment on tests/test_cli.py:1965) would lock this claim down.

src/lab_notebook/complete.py
```diff
@@ -0,0 +1,188 @@
...
+    the column, so a name that made it into the table is safe here). For a
+    `list` field, values are stored as JSON arrays; we json.loads and flatten in
+    Python (no json_each dependency). Any other type reads distinct raw values.
+    """
+    d = _dir()
+    if d is None:
+        return []
+    try:
+        schema = load_schema(d)
+    except LnbError:
```
carbonscott, line 119: This `except LnbError: return []` (and the mirror in `_schema`) is the "schema is missing" degradation README:231 promises, but no test reaches it: `test_unresolved_dir_returns_empty` fails earlier in `get_notebook_dir` (dir unresolved), so `_dir()` returns None and `load_schema` is never called. Consider a case where the dir resolves but `schema.yaml` is absent, asserting `--type` / `-f` / `-f val` all return `[]`.

src/lab_notebook/complete.py
```diff
@@ -0,0 +1,188 @@
...
+        return SUBCOMMANDS
+    sub = words[1]
+    cur = words[cword] if cword < len(words) else ""
+    prev = words[cword - 1] if cword >= 1 else ""
+
+    # --- value slots first (before flag-name) ---
+    # `-f <field> = <cur>`: '=' is a bash COMP_WORDBREAK, so key/=/value are
+    # separate words. Two shapes: prev=='=' (value being typed) or cur=='='
+    # (empty value right after `key=`).
+    if prev == "=" and cword >= 3 and words[cword - 3] in ("-f", "--field"):
```
carbonscott: Both value-slot shapes accept `--field` as well as `-f`, but the tests only exercise `-f` for values (`test_field_value_both_shapes`); the `--field` long form is covered only in the key slot. A single `--field repo = <TAB>` value assertion would cover this branch.

src/lab_notebook/complete.py
```diff
@@ -0,0 +1,188 @@
...
+# The subcommands a user can complete: the ten registered names plus
+# `completion` (the install helper). `__complete` is deliberately excluded — it
+# is an internal RPC target, not something a user types.
+SUBCOMMANDS = [
+    "init", "emit", "retract", "show", "sql", "search",
+    "schema", "rebuild", "contexts", "template", "completion",
+]
+
+# Per-subcommand flag names, offered in the flag-name slot (`<subcmd> -<TAB>`).
+FLAGS = {
```
carbonscott, line 35: Optional: `FLAGS` (and `SUBCOMMANDS`) are a second source of truth alongside the argparse subparsers in cli.py. They are in sync today (verified against the live parser), but a flag added to a subparser and forgotten here would silently stop completing with no failing test. A small test asserting these match the parser's subactions -- or deriving them from the parser -- would prevent quiet drift.

carbonscott (commented):
VERDICT: BLOCKING

## Review Summary

**Overall:** Clean, well-factored feature. The read-only guarantee is genuinely upheld in the code, the docs match behavior almost everywhere, and most completion slots are well-tested. One blocking gap: the headline "no-rebuild proof" is demonstrated only against a *fresh* index, where it cannot actually distinguish the read-only opener from the rebuild-capable path, and the explicitly-documented "tolerates a stale index" case has no test at all.

### Tests (blocking)
- `test_completion_never_rebuilds_existing_index` (tests/test_cli.py:1946) asserts `index.sqlite` mtime is byte-for-byte unchanged, but only against a *fresh* (non-stale) index. I verified locally that `ensure_db()` (the normal, rebuild-capable read path) *also* leaves a fresh index's mtime unchanged and prints nothing. So this "proof" would still pass even if `complete.py` were re-routed through `ensure_db` instead of `open_readonly` -- the mtime assertion is non-discriminating on a fresh index. The one place the two paths diverge is a *stale* index (schema.yaml newer than index.sqlite): `open_readonly` reads it untouched, while `ensure_db` rebuilds it (mtime changes + `Index rebuilt: N entries` on stderr). complete.py:14 and README:229 explicitly claim "tolerates a stale index" / "never rebuilds," yet no test exercises a stale index. Add a stale-index no-rebuild test (bump `schema.yaml` mtime past `index.sqlite`, exercise the index-backed slots, assert mtime unchanged and no "Index rebuilt") so the guarantee has real teeth. (Credit where due: `test_missing_index_returns_empty_and_creates_nothing` does have genuine teeth for the *creation* case -- it would catch a naive re-route to `ensure_db`.)

### Tests (non-blocking coverage gaps)
- Missing-*schema* degradation is untested. `_schema()` (complete.py:56) and `_field_values()` (complete.py:114-116) both swallow `LnbError` from `load_schema`, and README:231 claims "the schema is missing ... the affected slot simply offers nothing." But `test_unresolved_dir_returns_empty` fails earlier at `get_notebook_dir` (dir unresolved), so the "dir resolves but schema.yaml is absent" branch never actually runs. Add a case with a resolvable dir but no `schema.yaml`.
- `--field` long form is exercised only in the *key* slot (line 1886), never in the *value* slots. complete.py:173/175 accept `--field` in both `=`-wordbreak shapes, but only `-f` is tested for values (`test_field_value_both_shapes`). One `--field repo = <TAB>` value assertion would close it.

### Maintainability / drift (optional, recommended)
- `SUBCOMMANDS`/`FLAGS` in complete.py duplicate the argparse definitions in cli.py. I verified they are currently in sync with the live parser, but nothing guards against drift: a flag added to a subparser but omitted from `FLAGS` (or a new subcommand) would silently stop completing with no test failure -- exactly the kind of latent breakage worth a guard. Consider a test that asserts `SUBCOMMANDS`/`FLAGS` match the parser's subactions, or derive them from the parser. (Minor: `-h/--help` is intentionally omitted from every `FLAGS` entry -- fine, just noting it.)

### Docs (accurate)
- README's completion table, "schema-only slots like `--type` still work without an index," "never prints an 'Index rebuilt' line," and `open_readonly`'s `mode=ro`-not-`immutable` rationale all match the implementation. SKILL.md additions are accurate too. `test_complete_never_raises_on_garbage` (cword=99 genuinely raises `IndexError` inside `complete()`) is a nice, real exercise of the swallow path.

5 inline comments left on specific lines.

tests/test_cli.py
carbonscott, line 1979: This is the crux of the "no-rebuild proof," but it runs against a *fresh* index built by `_seed_completion`, and on a fresh index the assertion is non-discriminating. I verified that `ensure_db()` (the normal, rebuild-capable read path) also leaves a fresh index's mtime unchanged and prints nothing -- so this test would still pass even if the resolvers were re-routed from `open_readonly` to `ensure_db`. The two paths only diverge on a *stale* index (schema.yaml newer than index.sqlite): there `open_readonly` reads untouched while `ensure_db` rebuilds (mtime changes + "Index rebuilt" on stderr). Since complete.py:14 and README:229 explicitly promise "tolerates a stale index," please add a stale-index variant (bump schema.yaml mtime past the index, then exercise the index-backed slots and assert mtime unchanged + no "Index rebuilt"). That gives the guarantee teeth against a regression to the rebuild-capable path.

src/lab_notebook/complete.py
carbonscott, line 14: "tolerates a stale index" is the one read-only behavior with no test. It is genuinely true in the code (`open_readonly` never checks staleness and never routes to `ensure_db`), but the only no-rebuild test uses a fresh index where the rebuild-capable path is indistinguishable. A stale-index test (see comment on tests/test_cli.py:1965) would lock this claim down.

src/lab_notebook/complete.py
carbonscott, line 119: This `except LnbError: return []` (and the mirror in `_schema`) is the "schema is missing" degradation README:231 promises, but no test reaches it: `test_unresolved_dir_returns_empty` fails earlier in `get_notebook_dir` (dir unresolved), so `_dir()` returns None and `load_schema` is never called. Consider a case where the dir resolves but `schema.yaml` is absent, asserting `--type` / `-f` / `-f val` all return `[]`.

src/lab_notebook/complete.py
carbonscott: Both value-slot shapes accept `--field` as well as `-f`, but the tests only exercise `-f` for values (`test_field_value_both_shapes`); the `--field` long form is covered only in the key slot. A single `--field repo = <TAB>` value assertion would cover this branch.

src/lab_notebook/complete.py
carbonscott, line 35: Optional: `FLAGS` (and `SUBCOMMANDS`) are a second source of truth alongside the argparse subparsers in cli.py. They are in sync today (verified against the live parser), but a flag added to a subparser and forgotten here would silently stop completing with no failing test. A small test asserting these match the parser's subactions -- or deriving them from the parser -- would prevent quiet drift.

carbonscott (commented):
VERDICT: BLOCKING

## Review Summary

**Overall:** Clean, well-factored feature. The read-only guarantee is genuinely upheld in the code, the docs match behavior almost everywhere, and most completion slots are well-tested. One blocking gap: the headline "no-rebuild proof" is demonstrated only against a *fresh* index, where it cannot actually distinguish the read-only opener from the rebuild-capable path, and the explicitly-documented "tolerates a stale index" case has no test at all.

### Tests (blocking)
- `test_completion_never_rebuilds_existing_index` (tests/test_cli.py:1946) asserts `index.sqlite` mtime is byte-for-byte unchanged, but only against a *fresh* (non-stale) index. I verified locally that `ensure_db()` (the normal, rebuild-capable read path) *also* leaves a fresh index's mtime unchanged and prints nothing. So this "proof" would still pass even if `complete.py` were re-routed through `ensure_db` instead of `open_readonly` -- the mtime assertion is non-discriminating on a fresh index. The one place the two paths diverge is a *stale* index (schema.yaml newer than index.sqlite): `open_readonly` reads it untouched, while `ensure_db` rebuilds it (mtime changes + `Index rebuilt: N entries` on stderr). complete.py:14 and README:229 explicitly claim "tolerates a stale index" / "never rebuilds," yet no test exercises a stale index. Add a stale-index no-rebuild test (bump `schema.yaml` mtime past `index.sqlite`, exercise the index-backed slots, assert mtime unchanged and no "Index rebuilt") so the guarantee has real teeth. (Credit where due: `test_missing_index_returns_empty_and_creates_nothing` does have genuine teeth for the *creation* case -- it would catch a naive re-route to `ensure_db`.)

### Tests (non-blocking coverage gaps)
- Missing-*schema* degradation is untested. `_schema()` (complete.py:56) and `_field_values()` (complete.py:114-116) both swallow `LnbError` from `load_schema`, and README:231 claims "the schema is missing ... the affected slot simply offers nothing." But `test_unresolved_dir_returns_empty` fails earlier at `get_notebook_dir` (dir unresolved), so the "dir resolves but schema.yaml is absent" branch never actually runs. Add a case with a resolvable dir but no `schema.yaml`.
- `--field` long form is exercised only in the *key* slot (line 1886), never in the *value* slots. complete.py:173/175 accept `--field` in both `=`-wordbreak shapes, but only `-f` is tested for values (`test_field_value_both_shapes`). One `--field repo = <TAB>` value assertion would close it.

### Maintainability / drift (optional, recommended)
- `SUBCOMMANDS`/`FLAGS` in complete.py duplicate the argparse definitions in cli.py. I verified they are currently in sync with the live parser, but nothing guards against drift: a flag added to a subparser but omitted from `FLAGS` (or a new subcommand) would silently stop completing with no test failure -- exactly the kind of latent breakage worth a guard. Consider a test that asserts `SUBCOMMANDS`/`FLAGS` match the parser's subactions, or derive them from the parser. (Minor: `-h/--help` is intentionally omitted from every `FLAGS` entry -- fine, just noting it.)

### Docs (accurate)
- README's completion table, "schema-only slots like `--type` still work without an index," "never prints an 'Index rebuilt' line," and `open_readonly`'s `mode=ro`-not-`immutable` rationale all match the implementation. SKILL.md additions are accurate too. `test_complete_never_raises_on_garbage` (cword=99 genuinely raises `IndexError` inside `complete()`) is a nice, real exercise of the swallow path.

5 inline comments left on specific lines.

src/lab_notebook/complete.py
```diff
@@ -0,0 +1,188 @@
...
+
+    `words[0]` is the program name (`lab-notebook`); `cword` indexes the word
+    being completed. Never raises: unresolvable notebook / missing schema /
+    missing index all yield [].
+    """
+    if cword <= 1:
+        return SUBCOMMANDS
+    sub = words[1]
+    cur = words[cword] if cword < len(words) else ""
+    prev = words[cword - 1] if cword >= 1 else ""
```
carbonscott: The docstring above promises this function *never raises* ("unresolvable notebook / missing schema / missing index all yield []"), but neither `sub = words[1]` nor `prev = words[cword - 1]` has an upper-bound guard. `cur` is guarded with `if cword < len(words) else ""`; `prev` is guarded only on `cword >= 1`, never on `cword - 1 < len(words)`. Proven:

```
>>> complete(['lab-notebook','emit'], 7)   # cword past end -> prev = words[6]
IndexError: list index out of range
>>> complete(['lab-notebook'], 2)          # cword>=2, len(words)<2 -> sub = words[1]
IndexError: list index out of range
```

Not reachable through the shipped bash function (COMP_CWORD is always a valid index into COMP_WORDS), so non-blocking. But a direct caller relying on the contract gets a traceback — and `test_complete_never_raises_on_garbage` (cword=99) only passes because `cmd___complete` swallows the exception, i.e. the test is asserting the CLI net, not this contract. Direction: either guard `prev`/`sub` the same way `cur` is guarded (bounds-check `cword`), or soften the docstring so it stops promising something the code doesn't deliver.

carbonscott (commented):
VERDICT: LGTM

## Review Summary

**Overall:** A clean, well-scoped feature. The one-way `schema <- store <- complete <- cli` layering, the strictly read-only `open_readonly` (with a genuinely thoughtful `mode=ro`-not-`immutable=1` rationale), the SQL-injection guard that requires `field in schema["fields"]` before interpolating the column name, and the 16 tests covering every slot plus graceful degradation and a no-rebuild mtime proof are all excellent. Docs updated in the same PR per convention. No merge-blocking correctness bug in the reachable (shipped bash script) path.

That said, an adversarial pass on the slot logic turned up three real defects. None break documented, reachable behavior, so they are non-blocking — but two of them are only invisible because `cmd___complete` swallows *all* exceptions, and one directly contradicts a stated docstring contract. Worth addressing so the swallow backstops already-safe code rather than hiding defects.

### Key Findings (functionality / robustness)

1. **`complete()` violates its own "Never raises" docstring** (complete.py). `sub = words[1]` and `prev = words[cword - 1]` index without an upper-bound guard, unlike `cur`. `complete(['lab-notebook','emit'], 7)` and `complete(['lab-notebook'], 2)` both raise `IndexError`. Not reachable through the shipped bash function (COMP_CWORD is always a valid index into COMP_WORDS), but a direct caller relying on the contract — including the intent behind `test_complete_never_raises_on_garbage`, which only passes because the CLI layer swallows — gets a traceback.

2. **Value slots are not subcommand-scoped** (complete.py 178-183). They key only on `prev`, so `sql --type <TAB>` offers entry types, `retract --context <TAB>` offers contexts, and `show -f <TAB>` offers field keys — for flags those subcommands don't have. The flag-*name* slot right below (`FLAGS.get(sub, [])`) is correctly sub-scoped, so the two halves disagree. Nit: only misfires on invalid flag/subcommand combos.

3. **Corrupt index raises an uncaught `DatabaseError`** (complete.py 96, 146). The resolvers catch `sqlite3.OperationalError`, but a corrupt / non-database `index.sqlite` raises `sqlite3.DatabaseError` ("file is not a database"), which is the *parent* of `OperationalError`, not a subclass — so it escapes. The "tolerates a stale index" promise doesn't extend to a corrupt one; again masked only by the CLI swallow.

All three are proven with actual runs in the inline comments.

### PR Description Feedback

Description is thorough and clear — problem, design rationale, per-file breakdown, testing, and a documented known limitation. No notes.

4 inline comments left on specific lines.

src/lab_notebook/complete.py
```diff
@@ -0,0 +1,188 @@
...
+        return _field_values(words[cword - 2])
+    if cur == "=" and cword >= 2 and words[cword - 2] in ("-f", "--field"):
+        return _field_values(words[cword - 1])
+
+    if prev == "--type":
+        return _types()
+    if prev == "--context":
+        return _contexts()
+    if prev in ("-f", "--field"):
+        return _field_keys()   # each ends with "="
```
carbonscott: Consider: these value slots key only on `prev`, never on `sub`, so they fire for subcommands that have no such flag. Proven against a seeded notebook:

```
complete(['lab-notebook','sql','--type',''], 3)        -> ['observation','decision','dead-end','question','milestone']
complete(['lab-notebook','retract','--context',''], 3) -> ['alpha/one','beta/two']
complete(['lab-notebook','show','-f',''], 3)           -> ['repo=','branch=','tags=','artifacts=']
```

The flag-name slot just below is sub-scoped (`FLAGS.get(sub, [])` — e.g. `sql -<TAB>` correctly returns `[]`), so the two halves of the dispatcher disagree about whether the subcommand matters. It only misfires when the user has typed a flag invalid for that subcommand, so this is a nit — but gating these on `sub` (or on membership in `FLAGS[sub]`) would make the behavior consistent and avoid suggesting values for flags argparse will reject.

src/lab_notebook/complete.py
```diff
@@ -0,0 +1,188 @@
...
+                    values.update(str(v) for v in parsed)
+                else:
+                    values.add(str(parsed))
+            return sorted(values)
+        rows = conn.execute(
+            f"SELECT DISTINCT {col} FROM entries "
+            f"WHERE {col} IS NOT NULL ORDER BY 1"
+        ).fetchall()
+        return [str(r[0]) for r in rows]
+    except sqlite3.OperationalError:
```
carbonscott: Optional: this `except` (and the identical one in `_contexts`, line 96) catches only `sqlite3.OperationalError`. A corrupt / non-database `index.sqlite` raises `sqlite3.DatabaseError` ("file is not a database"), which is the *parent* class of `OperationalError`, not a subclass, so it escapes. Proven by writing garbage to `index.sqlite`:

```
complete(['lab-notebook','emit','--context',''], 3)  -> DatabaseError: file is not a database
complete(['lab-notebook','emit','-f','repo','='], 4) -> DatabaseError: file is not a database
# issubclass(sqlite3.DatabaseError, sqlite3.OperationalError) -> False
```

The module docstring says completion "tolerates a stale index" — but not a corrupt one; this path is masked only by the `cmd___complete` swallow. Catching `sqlite3.DatabaseError` (or `sqlite3.Error`) in both resolvers would make the read-only degradation actually cover a bad index file.

src/lab_notebook/cli.py
```diff
@@ -301,6 +330,23 @@ def cmd_template(args: argparse.Namespace) -> None:
...
+
+def cmd___complete(args: argparse.Namespace) -> None:
+    # Internal RPC target for the shell function. Must never raise on odd input
+    # (a Tab press should never spew a traceback), so swallow everything and
+    # emit nothing on failure. complete() already returns [] for the expected
+    # degradation paths; this also guards genuinely malformed argv.
+    try:
+        for candidate in complete(args.words, args.cword):
+            print(candidate)
+    except Exception:
```
carbonscott, line 357: FYI: as a final safety net for a Tab-time RPC, swallowing everything here is the right call — a Tab press must never spew a traceback. Flagging only because it is currently the *sole* guard preventing the `IndexError` (bad `cword`) and `DatabaseError` (corrupt index) paths in `complete.py` from surfacing: `complete()`'s "never raises" contract and the resolvers' error handling are both incomplete, so this `except Exception` is hiding real defects rather than backstopping already-safe code. Worth tightening the upstream guards (see the two `complete.py` comments) so this stays belt-and-suspenders.

src/lab_notebook/cli.py
```diff
@@ -61,6 +63,44 @@
...
+# `repo=` intact. The `cur=="="` remap covers the empty-value case
+# (`-f repo=<TAB>`). Passing `--` before the words makes argparse treat
+# flag-like tokens (`--type`, `-f`) as positionals rather than options.
+BASH_COMPLETION_SCRIPT = r"""_lab_notebook_complete() {
+    local cur="${COMP_WORDS[COMP_CWORD]}"
+    [[ "$cur" == "=" ]] && cur=""          # '=' wordbreak: empty value after key=
+    local line
+    COMPREPLY=()
+    while IFS= read -r line; do
+        [[ "$line" == "$cur"* ]] && COMPREPLY+=("$line")
```
carbonscott, line 93: The core of the fix, and it holds up under attack. `"$cur"` is quoted so its contents are a literal prefix (only the trailing unquoted `*` is a wildcard), and `COMPREPLY+=("$line")` is quoted so a candidate never word-splits or globs. I verified end-to-end that `$(...)`, backticks, and `*` from notebook data are inserted as inert text. Nice clean idiom that also works on bash 3.2. Nothing to change.

src/lab_notebook/cli.py
```diff
@@ -61,6 +63,44 @@
...
+    done < <(lab-notebook __complete "$COMP_CWORD" -- "${COMP_WORDS[@]}" 2>/dev/null)
+    # Field-key candidates end with '=' — keep the cursor attached so the value
+    # can be typed immediately. Only suppress the trailing space in that case.
+    if [[ ${#COMPREPLY[@]} -gt 0 && "${COMPREPLY[0]}" == *= ]]; then
```
carbonscott, line 97: FYI: this branch is reachable with an attacker-controlled candidate that ends in `=` (e.g. a context value `foo=`), but it's harmless — the candidate is never passed to `compopt`; only `-o nospace` is, so there's no injection surface and the effect is purely cosmetic (suppressed trailing space). No action needed.

src/lab_notebook/cli.py
```diff
@@ -61,6 +63,44 @@
...
+# (`-f repo=<TAB>`). Passing `--` before the words makes argparse treat
+# flag-like tokens (`--type`, `-f`) as positionals rather than options.
+BASH_COMPLETION_SCRIPT = r"""_lab_notebook_complete() {
+    local cur="${COMP_WORDS[COMP_CWORD]}"
+    [[ "$cur" == "=" ]] && cur=""          # '=' wordbreak: empty value after key=
+    local line
+    COMPREPLY=()
+    while IFS= read -r line; do
+        [[ "$line" == "$cur"* ]] && COMPREPLY+=("$line")
+    done < <(lab-notebook __complete "$COMP_CWORD" -- "${COMP_WORDS[@]}" 2>/dev/null)
```
carbonscott, line 94: FYI: because candidates are newline-delimited, a candidate that itself contains a literal newline splits into two `COMPREPLY` entries. It's benign (both halves are still assigned literally, so no execution) and contexts/tags/fields are newline-free in practice, so this isn't worth fixing — noting it only so the line-delimited protocol's one edge is on record.

src/lab_notebook/complete.py
```diff
@@ -0,0 +1,205 @@
...
+    except LnbError:
+        return []
+    fields = schema["fields"]
+    if field not in fields:
+        return []
+    conn = open_readonly(d)
+    if conn is None:
+        return []
+    try:
+        col = f'"{field}"'
```
carbonscott, line 128: FYI: the `if field not in fields` guard on line 122 correctly gates this identifier interpolation, and it matches the existing `build_sql` trust model (a field name that couldn't be a safe column would already have failed table creation). So this is not a new injection vector — the only actor who could abuse it already owns `schema.yaml` and the whole notebook. Fine as-is; flagging for completeness.

carbonscott (commented):
VERDICT: LGTM

## Adversarial re-review of fix `ef48241` (Tab-time RCE)

I treated the rewrite as guilty until proven innocent and drove the **real** generated bash function against the **real** CLI with a notebook seeded with hostile data. The RCE is genuinely closed and I found no new issue that the fix introduced. Suite: **162 passed**.

### 1. Original RCE — CLOSED (verified end-to-end)
Seeded a live notebook with contexts `$(touch SENTINEL)`, `` `touch SENTINEL` ``, and `*`, built the index, then sourced the emitted `completion bash` function and simulated `emit --context <TAB>` in a dir containing decoy files.
- No sentinel file was created.
- All three candidates landed in `COMPREPLY` **literally** (`$(...)` unexecuted, backticks unexecuted, `*` did **not** glob into the decoys).
- Teeth check: I ran the *old* `compgen -W` idiom against the identical candidates — it created the sentinel and expanded `*` into `decoy_a.txt decoy_b.txt`. So `test_bash_completion_inserts_candidates_literally` would fail on revert. Real teeth.

### 2. New break vectors against the `while read` idiom — all safe
- **Embedded newline in a candidate**: splits into two `COMPREPLY` entries (inherent to the line-delimited protocol). Both halves are still assigned literally — no execution. Benign; see FYI inline.
- **Leading/trailing whitespace + tab**: preserved verbatim by `IFS= read -r` (no stripping), inserted literally.
- **Leading `-` candidate** (`-rf`, `--danger`): assigned literally; `+=("$line")` does not treat it as an option.
- **Shell-meta candidates** (`;`, `|`, `&`, `>`, `$()`, backticks, `*`): all literal, no execution.
- **Malicious `cur`**: `[[ "$line" == "$cur"* ]]` treats `$cur` as a *literal* prefix (quoted) with only the trailing `*` as wildcard. `cur='*'` -> 0 matches (not match-all); `cur='a['` -> no bracket-expression error (matches only `a[x]`); backslash and `$(touch ...)` in `cur` -> no crash, no execution.
- **`compopt -o nospace` branch**: reachable with an attacker candidate ending in `=`, but the candidate is never passed to `compopt` (only `-o nospace` is), so there is nothing to inject; effect is cosmetic (suppressed trailing space).

### 3. Behavioral fixes — confirmed by real calls
- `complete(['lab-notebook','sql','--type'],3)` -> `[]` (value slots are subcommand-scoped); same for `retract --context`, `show -f`.
- `complete(['lab-notebook','emit'],7)` -> `[]` and `complete(['lab-notebook'],2)` -> `[]`, neither raises (`complete()` is bounds-total; the `except Exception` in `cmd___complete` is defense-in-depth, not load-bearing).
- Corrupt `index.sqlite` (garbage bytes): `_contexts()` and `_field_values('repo')` both return `[]` (the `sqlite3.Error` swallow catches the `DatabaseError`); no connection leak (try/finally close).
- `test_completion_does_not_rebuild_stale_index` has teeth: it forces staleness (schema newer than index), shows completion reads it untouched, then shows the normal path *does* rebuild the same index.

### Verdict
The fix does what it claims. Read-only, side-effect-free, candidates inert. No new vulnerability introduced. Only FYIs below, nothing blocking. (Posting as COMMENT rather than APPROVE because the reviewer account is the PR author.)

src/lab_notebook/cli.py
```diff
@@ -388,6 +445,23 @@ def main() -> None:
                             help="Overwrite existing schema.yaml")
     p_template.set_defaults(func=cmd_template)
 
+    # -- completion -- (prints a shell registration script)
+    p_completion = sub.add_parser("completion", help="Print a shell completion script")
+    p_completion.add_argument("shell", choices=["bash"])
+    p_completion.set_defaults(func=cmd_completion)
+
+    # -- __complete -- (hidden RPC target for the shell function; no help= so it
+    # stays out of the visible subcommand listing)
```
carbonscott: Nit: this overstates what omitting `help=` achieves. `lab-notebook --help` still lists `__complete` inside the `{init,...,completion,__complete}` choices metavar (both the usage line and the positional-arguments header) -- only the indented description line is suppressed. So `__complete` is not actually hidden from `--help`; the README's "hidden `__complete` subcommand" phrasing (L239) inherits the same imprecision. Not blocking, and argparse has no first-class way to drop one choice from the metavar. Consider softening to something like "omitted from the help *descriptions*" so the comment matches reality.

carbonscott (commented):
VERDICT: LGTM

## Review Summary

**Overall:** The reworked completion backend is correct and thoroughly tested. I verified every documented slot live against a seeded notebook, drove the real bash function end-to-end (including the `=` wordbreak, prefix filtering, and `nospace`), and confirmed the read-only guarantee. 162 tests pass. Nothing blocking; one comment-accuracy nit.

### What is genuinely good
- **RCE fix is correct.** `compgen -W` -> `while read` literal insertion + in-shell quoted prefix match. `test_bash_completion_inserts_candidates_literally` drives real bash against a poisoned backend and proves `$(...)`, backticks, and `*` stay inert text. I re-ran the equivalent end-to-end against a live poisoned notebook: candidates offered verbatim, nothing executes, `*` does not glob.
- **`complete()` is now genuinely total.** Confirmed out-of-range `cword`, empty `words`, and `cword>1` with no subcommand all yield `[]` *without* leaning on `cmd___complete`'s blanket `except`.
- **Subcommand-scoping the value slots introduced NO regression.** I exercised every path (subcommand / `--type` / `--context` / `-f` key / `key=value` in both wordbreak shapes with `-f` AND `--field` / flag-name) against `emit` and `search` -- all match the README behavior table -- and confirmed the mismatch cases (`sql --type`, `retract --context`, `show -f repo=`) correctly return `[]`. Live results: types, contexts (sorted+deduped), field keys (`repo= branch= tags= artifacts=`), list-field flattening (`tags=` -> `a b c`), and value prefix-filtering (`repo=ot` -> `other`) all behave as documented.
- **Stale-index test has real teeth.** It makes the index stale (schema newer than index -- the exact `ensure_db` rebuild trigger), proves completion reads it untouched (mtime byte-for-byte, no "Index rebuilt"), then shows the normal read path *would* rebuild it. This genuinely distinguishes read-only from rebuild-capable.
- **`build_parser()` extraction is behaviorally identical** to the old inline `main()`; no import-ordering issue (`complete` imports only `schema`/`store`, no cycle with `cli`).
- **Docs match code.** README table and skill/SKILL.md claims all check out, and the same-PR docs convention is honored.

### Drift-guard test -- assessed as robust
`test_completion_lists_match_parser` derives truth from the live parser and WOULD fail loudly (not silently no-op) on drift: adding a flag to a FLAGS-listed subparser breaks that sub's set-equality; adding a flag to a no-FLAGS subcommand breaks the final loop; adding/removing a subcommand breaks `set(SUBCOMMANDS) == user_subs`. If argparse internals ever changed, `assert len(subactions) == 1` fails rather than passing vacuously. Only (acceptable) gaps: it compares *sets*, so flag display order is not guarded, and it does not assert that a new value-flag also gets a resolver.

### Nit (non-blocking)
- `cli.py` ~L454 + `README.md` L239: the comment says omitting `help=` keeps `__complete` "out of the visible subcommand listing," but `lab-notebook --help` still prints `__complete` inside the `{init,...,completion,__complete}` choices metavar (both the usage line and the positional-arguments header). Only the indented *description* line is suppressed, so `__complete` is not fully hidden. See inline comment.

1 inline comment left. No blocking issues -- clear to merge.

PR #32, opened 2026-07-04
feat: US-010 - show with no id lists all entries

carbonscott:
## What

`lab-notebook show` now accepts an **optional** id.

- **`show <id>`** — unchanged: prints the one entry in full (key/value layout).
- **`show`** (no id) — lists **every live entry** as a compact table, newest first:

```
id|ts|context|type|content
20260704T030217-59d78649|2026-07-04T03:02:17+00:00|proj/c|dead-end|charlie third
20260704T030217-ca5117b3|2026-07-04T03:02:17+00:00|proj/b|decision|bravo second
20260704T030217-450c3ff4|2026-07-04T03:02:17+00:00|proj/a|observation|alpha first

(3 rows)
```

## How

- `show`'s positional `id` becomes `nargs="?", default=None`.
- `cmd_show` branches on `args.id is None`, running
  `SELECT id, ts, context, type, substr(content,1,80) AS content FROM entries ORDER BY ts DESC, rowid DESC`
  and reusing the existing `print_table`.
- The `AS content` alias makes the header read `content` (not the raw `substr(...)` expression).
- The `rowid DESC` secondary sort keeps ordering stable when entries share a timestamp — verified by a live smoke test where three real emits landed in the same second and still sorted newest-first.
- Retracted entries are already absent from the index, so the listing is live-only for free.

## Design decisions (confirmed with the maintainer)

- **Format:** compact table (`id | ts | context | type | content`), reusing `print_table` for consistency with `sql`/`search`.
- **Order/bound:** newest first (`ORDER BY ts DESC`), **all** live entries, no `LIMIT`.

## Tests

`tests/test_cli.py` `TestShow` gains: lists-all, newest-first, excludes-retracted, empty-notebook (`(no results)`), and an id-still-works regression. Full suite: **167 passed**.

## Docs

`show [ID]` section in README, `show --help`, the module docstring usage block, and the `/lnb` SKILL.md all updated.

### DIFF
--- a/src/lab_notebook/cli.py
+++ b/src/lab_notebook/cli.py
@@ -240,8 +240,11 @@ def cmd_show(args: argparse.Namespace) -> None:
     # No id: list every live entry as a compact table, newest first. Retracted
     # entries are already absent from the index, so the listing is live-only.
     # The `AS content` alias makes the header read `content` rather than the
-    # raw `substr(...)` expression; the `rowid DESC` tiebreak keeps ordering
-    # stable (newest last-ingested first) when two entries share a timestamp.
+    # raw `substr(...)` expression. `ts` is second-resolution, so `rowid DESC`
+    # is a deterministic tiebreak for entries sharing a timestamp (not a true
+    # sub-second clock: same-second emits across writers order by ingest, which
+    # need not match wall-clock) — this just keeps the output stable and
+    # repeatable rather than planner-dependent.
     if args.id is None:
         nb = Notebook(get_notebook_dir())
         try:
--- a/src/lab_notebook/cli.py
+++ b/src/lab_notebook/cli.py
@@ -4,7 +4,7 @@ Usage:
     lab-notebook init [path] [--template NAME]
     lab-notebook emit --context X --type Y "content"
     lab-notebook retract ID --reason "why"
-    lab-notebook show ID
+    lab-notebook show [ID]
     lab-notebook sql "SELECT ..."
     lab-notebook search "query"
     lab-notebook schema
@@ -237,6 +237,25 @@ def cmd_retract(args: argparse.Namespace) -> None:
 
 
 def cmd_show(args: argparse.Namespace) -> None:
+    # No id: list every live entry as a compact table, newest first. Retracted
+    # entries are already absent from the index, so the listing is live-only.
+    # The `AS content` alias makes the header read `content` rather than the
+    # raw `substr(...)` expression; the `rowid DESC` tiebreak keeps ordering
+    # stable (newest last-ingested first) when two entries share a timestamp.
+    if args.id is None:
+        nb = Notebook(get_notebook_dir())
+        try:
+            cursor = nb.query(
+                "SELECT id, ts, context, type, substr(content, 1, 80) AS content "
+                "FROM entries ORDER BY ts DESC, rowid DESC"
+            )
+            print_table(cursor)
+        except sqlite3.OperationalError as e:
+            raise LnbError(f"SQL error: {e}")
+        finally:
+            nb.close()
+        return
+
     nb = Notebook(get_notebook_dir())
     try:
         entry = nb.get(args.id)
@@ -409,8 +428,11 @@ def build_parser() -> argparse.ArgumentParser:
     p_retract.set_defaults(func=cmd_retract)
 
     # -- show --
-    p_show = sub.add_parser("show", help="Print one entry in full by id")
-    p_show.add_argument("id", help="Id of the entry to show")
+    p_show = sub.add_parser(
+        "show", help="Print one entry in full by id, or list all entries if no id is given")
+    p_show.add_argument(
+        "id", nargs="?", default=None,
+        help="Id of the entry to show; omit to list every entry newest-first")
     p_show.set_defaults(func=cmd_show)
 
     # -- sql --
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -617,7 +617,7 @@ class TestSql:
 # ---------------------------------------------------------------------------
 
 
-def make_show_args(entry_id):
+def make_show_args(entry_id=None):
     return argparse.Namespace(id=entry_id)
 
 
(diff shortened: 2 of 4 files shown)


src/lab_notebook/cli.py
```diff
@@ -237,6 +237,25 @@ def cmd_retract(args: argparse.Namespace) -> None:
 
 
 def cmd_show(args: argparse.Namespace) -> None:
+    # No id: list every live entry as a compact table, newest first. Retracted
+    # entries are already absent from the index, so the listing is live-only.
+    # The `AS content` alias makes the header read `content` rather than the
+    # raw `substr(...)` expression; the `rowid DESC` tiebreak keeps ordering
+    # stable (newest last-ingested first) when two entries share a timestamp.
```
carbonscott: Optional (comment wording, not code): this slightly overstates the tiebreak. `rowid` reflects *ingest* order (`sorted(glob('*.jsonl'))`, then append order within a file), not wall-clock emission order. Timestamps are second-resolution, so for entries that share a `ts` across *different writers*, `ORDER BY ts DESC, rowid DESC` groups by writer-file in reverse rather than by true emission time — four same-second emits interleaved between two writers list as `B2, A2, B1, A1`, not the strict `A2, B2, A1, B1`, so the globally-newest emit is not guaranteed to be the top row.

The clause still gives a *deterministic* order (an improvement over `sql`/`search`, which have no tiebreak at all), so no code change is needed — just consider softening 'newest last-ingested first' to something like 'deterministic' so the comment doesn't promise strict chronological order the second-resolution data model can't provide.

tests/test_cli.py
```diff
@@ -699,6 +699,67 @@ def test_notebook_get_missing_raises(self, notebook):
...
+        cmd_emit(make_emit_args(content="second bravo"))
+        id_b = _last_id_in_file(notebook, "test-writer.jsonl")
+        cmd_emit(make_emit_args(content="third charlie"))
+        id_c = _last_id_in_file(notebook, "test-writer.jsonl")
+        capsys.readouterr()  # clear emit output
+
+        cmd_show(make_show_args())
+        out = capsys.readouterr().out
+        # newest-first: C emitted last sorts before B, which sorts before A.
+        assert out.index(id_c) < out.index(id_b) < out.index(id_a)
```
carbonscott, line 728: Optional (test coverage): the `rowid DESC` tiebreak is described as load-bearing, but this test can't actually distinguish it. Mutation-testing shows that dropping `, rowid DESC` (leaving `ORDER BY ts DESC`) keeps this test green — same-`ts` rows still come back newest-first only because the reverse scan of `idx_entries_ts` incidentally yields rowid-DESC for ties. If that index were ever dropped or changed, `ORDER BY ts DESC` alone would return tied rows in rowid-ASC (oldest-first) order, silently inverting the contract with no failing test.

Consider a test whose pass/fail genuinely depends on the explicit tiebreak (e.g. construct two rows with equal `ts` and assert the higher-rowid one prints first), or at least a comment noting the tiebreak is only incidentally covered here.

tests/test_cli.py
```diff
@@ -699,6 +699,67 @@ def test_notebook_get_missing_raises(self, notebook):
...
+        cmd_emit(make_emit_args(content="drop this entry"))
+        drop_id = _last_id_in_file(notebook, "test-writer.jsonl")
+        cmd_retract(make_retract_args(drop_id))
+        capsys.readouterr()  # clear emit/retract output
+
+        cmd_show(make_show_args())
+        out = capsys.readouterr().out
+        assert keep_id in out
+        assert drop_id not in out
+        assert "(1 row)" in out
```
carbonscott, line 742: Nice — asserting the exact `(1 row)` count pins retracted-exclusion independently of the substring checks, and pairing it with the full-8-hex-id `drop_id not in out` makes an accidental substring match effectively impossible. This is a strong, false-positive-resistant assertion.

carbonscott (commented):
## Review Summary

**Overall:** Clean, well-scoped change that makes `lab-notebook show` list all entries when no id is given. It improves code health and I'd approve it. (Posted as **COMMENT** rather than APPROVE only because the reviewer and author are the same GitHub account — GitHub blocks self-approval.)

Reviewed adversarially with two independent passes: a correctness hunt (running the CLI against hand-built notebooks) and mutation-testing of the new tests. No blocking issue survived.

### What's good
- Reuses `print_table` and the existing `entries` index rather than inventing a new path — consistent with `sql`/`search`, and retracted entries are excluded for free.
- The `AS content` alias so the header reads `content` (not the raw `substr(...)`) is a nice touch, and it's pinned by an exact-header assertion.
- The tests are genuinely strong: mutation-testing confirmed each new test goes red for the defect it targets (ordering, header/column set, empty-state `(no results)`, id-vs-list branch). `test_no_arg_excludes_retracted` pins the exact `(1 row)` count, so exclusion is verified independently of the content check.
- Docs updated in all the right places (README `show [ID]`, `--help`, module docstring, `/lnb` SKILL.md).

### Findings (all non-blocking)
- **Ordering comment overstates the tiebreak** (`cli.py`): `rowid` is *ingest* order, not wall-clock. For same-second emits across different writers, the newest emit is not guaranteed to be the top row. The clause still yields a deterministic order (better than `sql`/`search`, which have none), so this is a comment-wording fix, not a code fix. Detail inline.
- **The `rowid DESC` tiebreak is only incidentally tested**: dropping it keeps the suite green, because the reverse scan of `idx_entries_ts` happens to return ties in rowid-DESC order. Load-bearing behavior that no test actually pins. Detail inline.
- **Nit:** the 80-char `substr` cap is not asserted by any test (widening it to unbounded stays green). Very minor.

### Checked and fine (not issues)
- CSV `|`/quote/newline/unicode handling is identical to `sql`/`search` (shared `print_table`) — no new corruption, and the `(N rows)` footer counts DB rows correctly even when a field wraps.
- Empty notebook, retracting the only entry, empty/whitespace content, and unset `LAB_NOTEBOOK_DIR` all produce clean messages with no tracebacks.
- Rebuild preserves ordering; tombstones never leak into the listing.

3 inline comments below.

PR #33, opened 2026-07-04
feat(find): JSONL output via lnb find -o json

carbonscott:
## Summary

Adds an opt-in `lnb find -o json` / `--output json` mode emitting machine-readable **JSONL** — one compact JSON object per matched live entry, one per line. It mirrors the on-disk `.lnb/*.jsonl` shape: `find -o json` reads as "a `cat` that resolves filters + retractions + dedup." Records are emitted verbatim (extras stay as top-level typed keys), so it is lossless where the existing `sql` TSV path is not.

Scope was decided by a **unanimous 3-lens persona GO** (Karpathy / geohot / Conway): `find` **only**. `note`, `sql`, and `retract` deliberately do **not** get `-o json` (see `.goal/lnb-json-output.spec.md` for the 11-point design contract and rationale).

## What changed

- `parse()` widened to carry an `output` value and accept `-o` / `--output`.
- New small `emit_json()` JSONL helper: `print(json.dumps(record, ensure_ascii=False))` per record.
- `cmd_find` selects the emitter and **drops the single-id human DWIM in json mode** — a unique id-fragment match emits ONE plain object; zero matches emit zero lines at exit 0.
- **stdout carries only JSONL; every diagnostic goes to stderr** and never interleaves.
- No 80-char content truncation in json mode (that truncation is display-only).
- Live-only: retracted entries and `_retract` tombstones never appear.
- **Fail closed on misuse:** `find` rejects an unknown `-o` value loudly (naming `json`); `note` and `retract` reject `-o`/`--output` rather than silently accepting-and-no-op'ing.

## Testing

- `python3 -m pytest -q` → **41 passed, 0 failed** (11 new tests, all additive; existing 30 unweakened).
- `lnb find <ID> -o json | python3 -m json.tool >/dev/null; echo rc=$?` → **rc=0**.
- Multi-row `find -o json` validated as per-line JSONL (each line parses independently).
- Empty/no-match `find -o json` → zero stdout lines, exit 0, diagnostic on stderr.

## Minimality

- `wc -l lnb.py` ≈ **392** — still one file, stdlib-only (`json` already imported).
- `pyproject.toml` `dependencies = []` — no new runtime dependency.

## Scope excluded

The live PR-review orchestration files `.goal/lnb-json-pr-review.{json,ledger.json}` are intentionally **not** committed — they are the running contract/ledger of the current review loop, not part of the feature.

### DIFF
--- a/.goal/lnb-json-output.ledger.json
+++ b/.goal/lnb-json-output.ledger.json
@@ -1,0 +1,45 @@
+{
+  "goal": "On a new branch off dev-min in the minimal lnb worktree, decide via an independent persona debate whether `-o json` earns its place; if GO, implement it on the surfaces that earn it with an independent review after each change while staying minimal (single stdlib-only file); if NO-GO, record the reasoned rejection.",
+  "slug": "lnb-json-output",
+  "contract": ".goal/lnb-json-output.json",
+  "created": "2026-07-04T21:47:21+00:00",
+  "branch": "feat-json-output",
+  "limits": { "subagents_min": 2, "subagents_max": 5, "min_iterations": 3, "max_turns": 9 },
+  "iterations": [
+    {
+      "n": 1,
+      "thought": "Before writing any code, decide via an independent 3-lens persona debate whether `-o json` earns its place in this deliberately minimal tool, and on which command surfaces. Feed the debaters the real tensions (sql escape hatch already exists; the store is already JSONL on disk) so a NO-GO is a live option.",
+      "delegated": [
+        { "agent": "general-purpose (karpathy lens)", "task": "GO/NO-GO on `-o json` from Karpathy's minimal-single-file lens; name surfaces that earn it", "model": "opus-4.8 (inherited)" },
+        { "agent": "general-purpose (geohot lens)", "task": "GO/NO-GO from geohot's concept-count/fail-closed lens; is json a new concept or redundant surface?", "model": "opus-4.8 (inherited)" },
+        { "agent": "general-purpose (damian-conway lens)", "task": "GO/NO-GO from Conway's DWIM/diagnostics/edge-case lens; natural output-format seam or feature creep?", "model": "opus-4.8 (inherited)" }
+      ],
+      "integrated": "Unanimous 3/3 GO. All three name `find` as the surface that earns it (sole surface doing semantic work cat/sql can't reproduce losslessly: filter grammar + retraction-resolution + dedup). Unanimous NO on `sql` (already the machine hatch; SQLite has json_object()) and `retract` (one-line receipt). `note -o json` was a 2-1 rider (conway for, karpathy/geohot against) -> DECLINED for minimality (id already printed and on disk). Synthesized 11 unanimous design constraints into .goal/lnb-json-output.spec.md: JSONL (not array), ensure_ascii=False, records verbatim (extras top-level), invariant shape (no single-id DWIM object), no truncation, live-only, stdout=JSON/stderr=diagnostics, empty=0-lines-exit-0, opt-in only, reject unknown -o loudly, stdlib-only one file.",
+      "new": "DECISION: GO on `-o json`, scope = `find` ONLY (sql/retract/note excluded, note-rider considered and declined). New artifact: .goal/lnb-json-output.spec.md capturing the 11-point unanimous design contract.",
+      "remaining": "Implement `find -o json` per spec (JSONL emitter) + add pytest coverage for json output; independent review after the change with all changes-requested items resolved and re-reviewed; then harden edge cases (empty/no-match/malformed/extras/retracted/unknown-value), prove `python -m json.tool` rc=0 for a real entry, prove per-line JSONL validity for multi-row, and confirm minimality (wc -l lnb.py, pyproject deps unchanged)."
+    },
+    {
+      "n": 2,
+      "thought": "GO is decided and scoped to `find`. Implement `find -o json` strictly to the frozen spec, keep it minimal, and get an INDEPENDENT review of the actual change plus independent json test coverage -- all as separate subagents against real code, not self-review.",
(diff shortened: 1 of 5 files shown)


lnb.py
```diff
@@ -301,6 +311,11 @@ def cmd_sql(args):
 
 # --- rendering, diagnostics & helpers ---------------------------------------
 
+def emit_json(rows):
```
carbonscott, line 314: FYI (nice): this is a clean, minimal, lossless emitter — one verbatim record per line with `ensure_ascii=False` matching `append()`'s on-disk encoding, so `find -o json` really does read as "a `cat` that resolves filters + retractions + dedup." And because `json.dumps` escapes embedded newlines, a multi-line `content` stays a single physical line, so per-line JSONL parsing stays valid. Good design.

lnb.py
```diff
@@ -198,7 +202,11 @@ def cmd_note(args):
 
 
 def cmd_find(args):
-    terms, ctype, context, _ = parse(args)
+    terms, ctype, context, _, output = parse(args)
+    if output is not None and output != "json":
```
carbonscott, line 206: Consider (Documentation): this adds a user-facing `-o json` / `--output json` flag, but the `USAGE` string (lines 41–43) and the module docstring still show `find [terms] [@ctx] [+type]` with no mention of it, so `lnb --help` won't advertise the feature. A one-token addition to `USAGE` (e.g. `[-o json]` on the `find` clause) would make it discoverable. Non-blocking.

lnb.py
```diff
@@ -249,6 +257,8 @@ def cmd_retract(args):
         if a == "--reason":
             i += 1
             reason = args[i] if i < len(args) else None
+        elif a in ("-o", "--output"):
```
carbonscott, line 260: Consider: this correctly rejects the exact `-o`/`--output` tokens, but the loop is still fail-open for everything else — `--output=json` (equals form) or any unknown flag falls through to `elif target is None`, is silently ignored once `target` is set, and the retraction still proceeds. That's the inverse of `parse()`, which `die()`s on unexpected args, and of the module's stated fail-closed philosophy. An explicit trailing `else: die(...)` for unrecognized tokens would close the gap. Low practical risk, so non-blocking — flagging because this PR specifically hardened this loop.

carbonscott (commented):
## Review Summary

**Overall:** Clean, well-scoped, well-tested addition — an opt-in `find -o json` JSONL mode in a genuinely minimal diff (+26/-11 in `lnb.py`). No blocking correctness issues found; a few non-blocking notes below. Posting as a **COMMENT** (this is advisory review input, not an approval — and GitHub blocks a PR author from approving their own PR anyway).

### What's good
- `emit_json` is a two-line, lossless JSONL helper mirroring the on-disk shape; extras stay top-level typed keys (the `sql` TSV path can't do this). Because `json.dumps` escapes embedded newlines, multi-line content stays a single physical line, so JSONL integrity holds.
- Clean stdout/stderr discipline: stdout carries only JSONL; every diagnostic (`empty`, `no matches`, `N ids match`) goes to stderr and never interleaves.
- Fail-closed hardening: `note` and `retract` reject `-o/--output` loudly instead of silently no-op'ing; `find` rejects an unknown format naming the accepted vocabulary (`json`).
- Strong, purely additive test coverage (11 tests, 30→41) pinning the contract: invariant one-object shape, empty=0-lines-exit-0, retraction exclusion, no display truncation, `--output` long-form parity, and the `-o` missing-value case.

### Non-blocking findings
- **Documentation:** the new user-facing `-o json` flag isn't advertised in the `USAGE` string (lines 41–43) or the module docstring, so `lnb --help` won't surface it. (inline)
- **retract parser is fail-open for unknown tokens:** it rejects the exact `-o`/`--output` tokens, but `--output=json` (equals form) or any other unknown flag falls through and is silently ignored once `target` is set, and the retraction still proceeds — the inverse of `parse()`, which `die()`s on unexpected args. Low practical impact (retract is the user's explicit intent), but it leaves the fail-closed boundary partial. (inline)
- **Nit:** at the unique-id branch, `emit_json(hits)` is used directly while sibling paths use the `emit` binding. (inline)
- **Design / process hygiene:** the `.goal/*` orchestration files (contract + live ledger) are committed into this feature PR targeting `dev-min`. The ledger is running process state, not part of the JSONL feature. (inline)

### PR Description Feedback
The description is excellent — clear first-line summary, rationale, explicit scope decision, testing evidence, and minimality accounting. No changes needed.

_5 inline comments left._

lnb.py
```diff
@@ -214,10 +222,10 @@ def cmd_find(args):
     if len(terms) == 1 and not context and not ctype and ID_RE.match(terms[0]):
         hits = [e for e in rows if terms[0] in e.get("id", "")]
         if len(hits) == 1:
-            return show_entry(hits[0])
+            return emit_json(hits) if as_json else show_entry(hits[0])
```
carbonscott: Nit: `emit` is already bound to `emit_json` when `as_json` is true (line 209), so this could read `return emit(hits) if as_json else show_entry(hits[0])` for a single source of truth on the json emitter. The human branch genuinely needs `show_entry` (full vertical dump), not `emit`/`print_table`, so the ternary itself is correct — purely a consistency nit.

.goal/lnb-json-output.ledger.json
```diff
@@ -0,0 +1,45 @@
+{
```
carbonscott, line 1: Consider (mixed concerns): this commits the goal-loop's live process ledger (and the `.json` contract) into a feature PR targeting `dev-min`. The ledger is running orchestration state rather than part of the JSONL feature, so it adds process bookkeeping to the product history. `lnb-json-output.spec.md` is a reasonable design artifact to keep; the ledger/contract may be better left out of the merge (or in a separate docs/process location). Non-blocking.

PR #34, opened 2026-07-06
feat(lnb): find→log redesign — JSONL producer, jq is the consumer

carbonscott:
## What & why

Implements the converged **`find`→`log` redesign** (design-only PRD `design/lnb-jq-consumer-redesign.md` + the settled deltas in the handoff). The thesis: **lnb is a JSONL producer; `jq` is the consumer.** Every change *net-removes* a concept.

Verb set is now **`note` (append) · `log` (dump) · `retract` (tombstone)**; bare `lnb` = help.

## Changes (all net-subtraction)

- **Rename read verb `find` → `log`.** `log` is a literal, argument-free, **filter-free** emitter: it prints *every* record as JSONL — entries **and** `_retract` tombstones — uncapped, ascending oldest→newest via `sorted()`. All selection (content, context, type, **liveness**) is `jq`'s job. A stray arg to `log` is a fail-closed usage error.
- **Retire `lnb sql` + the `sqlite3` dependency.** The tool is now **stdlib-only** (`dependencies = []`).
- **Remove the 10-entry default-view cap** and its stderr side-channel.
- **Remove `-o json` as a mode** (JSONL is the only output); delete `print_table` / `show_entry`.
- **`scan()` → `(all_records_incl_tombstones ascending, retracted_set)`.** Liveness is no longer computed on read.
- **`retract` re-owns its liveness guard.** One `live` list (`type != "_retract"` **and** `id not in retracted`) feeds both hit branches and `suggest_id` — so retract never targets a tombstone or an already-dead id.
- **Ship one canonical live-view `jq` recipe in `--help`** (Amendment 1) — the producer teaching its consumer, at zero code cost.
- Delete now-orphaned `no_matches` / `vocab` / `ID_RE`; refresh `README.md` and `SKILL.md` (the SKILL read/write `log` collision resolved by hand).

`lnb.py`: **392 → 324 lines**, zero non-stdlib imports.

## The load-bearing correctness point

Because `scan()` now returns tombstones, `retract` can no longer lean on the read path having excluded them. A **partial** guard (checking only `id not in retracted`, forgetting `type != "_retract"`) passes the *entire* rest of the suite yet **still retracts a tombstone**. This is pinned by `test_retract_of_tombstone_id_is_no_entry`, which exercises the tombstone-type path *distinctly* from the already-retracted path. An independent verifier re-derived this from scratch: the test **fails** against a weakened guard and **passes** against this branch.

## Testing

- **50 passed** (`uv run pytest -q`), stdlib-only, jq recipe tests run against real `jq 1.7`.
- The live-view test extracts the recipe **from `--help`** and runs it — so the shipped copy can never silently drift from what's tested.
- End-to-end verified with real bytes: `note` a decision + a note, `retract` the decision → `lnb log` shows all three lines incl. the tombstone; the `--help` live-view idiom leaves only the note; the *naive* `jq 'select(.type=="decision")'` still returns the dead decision (the documented fail-open footgun).

## Notes

- Documented limitation (not engineered around): `scan()` sorts on the raw ISO ts **string**; string order ≠ instant order only across a *changing* UTC offset — harmless for the single-user, single-timezone target.
- Design/planning artifacts live under the git-ignored `.artifacts/` (persona plans, finalized design, red-team findings) — not part of this diff.

https://claude.ai/code/session_01KpwSWUqhDg9c7N7qKho4Ug

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: lnb
-description: "Record and recall research notebook entries. Use when the user asks to record an observation, decision, dead-end, question, or milestone, or asks what we've tried / decided / hit. Trigger on: 'note this', 'log this', 'record', 'what have we tried', 'what did we decide', 'search the notebook'."
+description: "Record and recall research notebook entries. Use when the user asks to record an observation, decision, dead-end, question, or milestone, or asks what we've tried / decided / hit. Trigger the WRITE (Note) action on: 'note this', 'log this', 'record'. Trigger the READ (Recall) action on: 'recall', 'what have we tried', 'what did we decide', 'search the notebook'."
 user-invocable: true
 argument-hint: "<note|recall> [text...]"
 ---
@@ -13,10 +13,15 @@ whole notebook as JSONL) — so at this skill layer "log" is never a write; use
 **note**. The CLI (`lnb`) auto-discovers the notebook (nearest `.lnb/` walking
 up, or `$LNB_DIR`) and creates one on the first `note`. No setup step.
 
-Parse the intent of `$ARGUMENTS`: a request to record something (first word
-`note`/`record`, or clearly a new observation/decision/dead-end) → **Note**; a
-request to recall (first word `recall`, or a question about what was tried /
-decided) → **Recall**; if ambiguous, show the two usages below.
+Parse the intent of `$ARGUMENTS`. A request to record something — first word
+`note`/`log`/`record`, a phrasing like "log this"/"note this", or clearly a new
+observation/decision/dead-end → **Note** (the WRITE action; runs `lnb note …`).
+A request to recall — first word `recall`, "what did we / have we …", or a
+question about what was tried / decided → **Recall** (the READ action; runs
+`lnb log | jq …`). If ambiguous, show the two usages below. Note the split:
+the user saying "log this" means WRITE an entry (→ **Note**), whereas the *CLI*
+verb `lnb log` READS the whole notebook (see above) — at this skill layer "log"
+is never the write command, `lnb note` is.
 
 ## Note
 
--- a/SKILL.md
+++ b/SKILL.md
@@ -1,20 +1,24 @@
 ---
 name: lnb
-description: "Log and recall research notebook entries. Use when the user asks to record an observation, decision, dead-end, question, or milestone, or asks what we've tried / decided / hit. Trigger on: 'log this', 'note this', 'record', 'what have we tried', 'what did we decide', 'search the notebook'."
+description: "Record and recall research notebook entries. Use when the user asks to record an observation, decision, dead-end, question, or milestone, or asks what we've tried / decided / hit. Trigger on: 'note this', 'log this', 'record', 'what have we tried', 'what did we decide', 'search the notebook'."
 user-invocable: true
-argument-hint: "<log|recall> [text...]"
+argument-hint: "<note|recall> [text...]"
 ---
 
 # Lab Notebook (`/lnb`)
 
-A git-tracked, append-only research notebook. Two verbs: **log** and **recall**.
-The CLI (`lnb`) auto-discovers the notebook (nearest `.lnb/` walking up, or
-`$LNB_DIR`) and creates one on first `log`. No setup step.
+A git-tracked, append-only research notebook. Two skill verbs: **note** (write)
+and **recall** (read). NB: the *CLI* read verb is now `lnb log` (it emits the
+whole notebook as JSONL) — so at this skill layer "log" is never a write; use
+**note**. The CLI (`lnb`) auto-discovers the notebook (nearest `.lnb/` walking
+up, or `$LNB_DIR`) and creates one on the first `note`. No setup step.
 
-Parse the first word of `$ARGUMENTS`: `log` → **Log**, `recall` → **Recall**,
-anything else → show the two usages below.
+Parse the intent of `$ARGUMENTS`: a request to record something (first word
+`note`/`record`, or clearly a new observation/decision/dead-end) → **Note**; a
+request to recall (first word `recall`, or a question about what was tried /
+decided) → **Recall**; if ambiguous, show the two usages below.
 
-## Log
+## Note
 
 Append one entry. **Distill first, then confirm, then write** — the log is
 append-only, so a wrong entry stays as history.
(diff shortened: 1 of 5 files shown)


lnb.py
```diff
@@ -270,90 +255,37 @@ def cmd_retract(args):
...
-    rows, retracted = scan(nbdir)
-    hits = [e for e in rows if e.get("id") == target] or \
-           [e for e in rows if target in e.get("id", "")]
+    records, retracted = scan(nbdir)
+    # scan() now returns tombstones and dead entries too, so retract must own
+    # its liveness: only REAL, still-live entries are retractable candidates.
+    # Both guards matter -- `type != "_retract"` stops a substring `target`
+    # from matching a tombstone's own id ("retract a tombstone"); `id not in
+    # retracted` stops a second tombstone for an already-dead id.
+    live = [e for e in records
```
carbonscott, line 268: This is the load-bearing correctness point, and it is implemented correctly: one `live` list guards on BOTH `type != "_retract"` and `id not in retracted`, feeds both hit branches (exact then substring), and `suggest_id(target, live)` receives the filtered list. I verified the RT1 edge - retracting a tombstone's own id or its random suffix falls through to `no entry` with no second tombstone appended - and that a partial guard dropping the `type` check would still substring-match the fresh tombstone (whose id is not yet in `retracted`). The guard is real, not decorative. Nice.

tests/test_lnb.py
```diff
@@ -447,173 +634,76 @@ def test_retract_already_retracted_reports_clearly(nb_env):
...
-    b = parse_noted(run_lnb(["note", f"second {marker}"], env=nb_env).stdout)["id"]
-    assert run_lnb(["retract", a, "--reason", "drop it"],
+    assert n_tombstones(nb_env) == 1
+    r2 = run_lnb(["retract", victim, "--reason", "second"], env=nb_env)
+    assert r2.returncode != 0
+    assert "already retracted" in r2.stderr.lower()
+    assert n_tombstones(nb_env) == 1, "no second tombstone for a dead id"
+
+
+def test_retract_of_tombstone_id_is_no_entry(nb_env):
```
carbonscott, line 718: FYI (positive): this is the highest-value pin and it is genuinely differential. Because a fresh tombstone's own id is not yet in `retracted`, a partial guard that checks only `id not in retracted` would substring-match the tombstone and append a second one at exit 0 - so this test fails against the weakened guard while passing here. Exercising both the full tombstone id and its random suffix covers the substring attack surface. Well constructed.

tests/test_lnb.py
```diff
@@ -20,6 +31,19 @@
...
+# IN the help text AND runs JQ_LIVE_FILTER through real jq, so the shipped copy
+# can never silently rot.
+JQ_LIVE_FILTER = (
+    'map(select(.type=="_retract").retracts) as $dead\n'
+    '      | .[] | select(.type != "_retract" and (.id | IN($dead[]) | not))'
+)
+RECIPE = "lnb log | jq -s '" + JQ_LIVE_FILTER + "'"
+
+HAVE_JQ = shutil.which("jq") is not None
+needs_jq = pytest.mark.skipif(not HAVE_JQ, reason="jq is not installed")
```
carbonscott, line 53: FYI: the two highest-value recipe pins (`test_live_view_jq_idiom_drops_retracted` and `test_naive_filter_keeps_retracted_footgun`) are gated on `needs_jq`. In an environment without `jq` they silently skip and the suite still reports all-green, leaving the shipped recipe's *correctness* unverified. That is a reasonable tradeoff for a stdlib-only tool, but worth ensuring CI always has `jq` on PATH (it does here: jq 1.7) so these do not degrade into no-ops.

SKILL.md
```diff
@@ -1,20 +1,24 @@
 ---
 name: lnb
-description: "Log and recall research notebook entries. Use when the user asks to record an observation, decision, dead-end, question, or milestone, or asks what we've tried / decided / hit. Trigger on: 'log this', 'note this', 'record', 'what have we tried', 'what did we decide', 'search the notebook'."
+description: "Record and recall research notebook entries. Use when the user asks to record an observation, decision, dead-end, question, or milestone, or asks what we've tried / decided / hit. Trigger on: 'note this', 'log this', 'record', 'what have we tried', 'what did we decide', 'search the notebook'."
```
carbonscott: Nit: the read/write `log` collision is well resolved (write action renamed to **Note**, with the explicit NB below). One residual: this trigger list still advertises `'log this'`, while the parse rule keys write-intent on first word `note`/`record`. A user who types `/lnb log ...` matches neither `note`/`record` nor `recall`, so routing falls back to the "clearly a new observation -> Note" / "ambiguous -> show usages" branch. Not blocking - consider adding `log`/`log this` explicitly to the Note-intent list (at this layer it means record), so the advertised trigger and the parse rule agree.

carbonscott (commented):
## Review Summary

**Verdict: APPROVE** (posted as a `COMMENT` review only because GitHub forbids approving one's own PR - there are no blocking findings; every inline comment is positive / FYI / Nit).

**Overall:** A clean, faithful, well-tested implementation of the finalized `find`->`log` redesign. Every change is net-subtraction, the load-bearing retract liveness guard is correct in every branch, and the test suite reads as a genuine executable contract. Independently reproduced: **50 passed** against real `jq 1.7`.

### Design
- The `note`/`log`/`retract` three-verb split, with `log` a filter-free JSONL emitter and `jq` as the consumer, is coherent and matches the stated thesis. Deleting `sql`, `-o json`, the 10-entry cap, and all read-path filters removes the advertised ~13 concepts with no capability regression (`jq` covers each), and `lnb.py` is now genuinely stdlib-only. Good subtraction.

### Functionality (the correctness-critical part)
- **The retract liveness guard is correct.** `scan()` now returns tombstones + dead entries, and `cmd_retract` re-owns liveness with one `live` list filtered on BOTH `type != "_retract"` AND `id not in retracted`, used by both hit branches (exact then substring) and passed to `suggest_id`. I verified the RT1 fail-open (retracting a tombstone's own id, and its random suffix) falls through to `no entry` and appends no second tombstone, and that the partial-guard trap (dropping the `type` check) is genuinely closed. Inline note on `lnb.py:264`.
- **All `scan()` consumers are correct.** Only `cmd_log` (wants the raw log verbatim) and `cmd_retract` (filters to `live`) read its return value; `cmd_sql` is deleted. No reader silently inherited the old "live answer" meaning.
- **The `--help` recipe is the exact copy the tests run.** `test_live_view_jq_idiom_drops_retracted` extracts `RECIPE` from `--help` and runs it through real `jq`, and `retracts` is pinned as a scalar id string (the recipe's load-bearing contract). The README/SKILL copies use a 2-space hanging indent vs the docstring's 6-space (its 2-space indent under the 4-space docstring block); functionally identical since `jq` ignores whitespace, and the *tested* copy is the `--help`/docstring one.

### Tests
- Differential, not gameable. `test_retract_of_tombstone_id_is_no_entry` exercises the tombstone-type path distinctly from already-retracted, so a weakened guard fails it. Deletions correspond to genuinely-removed behavior (sql, `-o json` mode, cap, read filters, id-lookup, `no_matches`) - no test was dropped to hide surviving behavior. No xfail. One legitimate `skipif(not jq)` - see inline FYI on `tests/test_lnb.py:45`.

### Documentation
- README and SKILL are fully refreshed: no stray `sql` / `find` / `-o json` / "10 most recent" remain. The SKILL read/write `log` collision is resolved - the write action is renamed to **Note**, "log" reserved for the CLI read verb, with an explicit NB. One residual nit inline on `SKILL.md:3`.

4 inline comments below (all positive / FYI / Nit - none blocking).

PR #35, opened 2026-07-07
perf(log): stream ts-ordered records via heapq.merge instead of a full in-memory sort

carbonscott:
## What

`lnb log`'s read path previously loaded **every** record into memory and called `sorted(...)`. This swaps that for a streaming k-way merge (`heapq.merge`) over the per-writer JSONL files, each already `ts`-ascending by construction (append-only, single writer per file). `retract` still uses the materialized `scan()`; **only the `log` read path changes**. On-disk JSONL format, CLI grammar, and everything else are untouched — no index, schema, or config added.

## Why — measured on a ~1M-record / 184 MB synthetic notebook

| Metric | Baseline (`sorted`) | Patched (`heapq.merge`) | Delta |
|---|---|---|---|
| **Peak RSS** | 1051 MB | 16 MB | **−98.5% (~65×)** |
| Wall-clock (`log \| wc -l`) | 6.12 s | 5.56 s | −9.2% |
| Time-to-first-line | ~2.4 s | ~23 ms | **~100× faster** |
| User CPU | ~5.48 s | ~5.41 s | ~unchanged |

Method: one warm-up run to prime the page cache, then the median of 3 runs under `/usr/bin/time -v` on the same CPython 3.12.3. The memory drop is the headline: peak RSS falls from ~5.5× the input size to a small constant, because the merge holds only one record per open file at a time instead of the whole dataset plus its sorted copy. User CPU is essentially unchanged, so the ~9% wall-clock gain comes from not allocating/GC-ing a 1M-element list, not from cheaper compute.

## Correctness — output is byte-identical, proven three ways

- `sha256` of the **full** `log` output is EQUAL between baseline and patched (`581e58c718eed3c4a45e8262624804a173d17923fc45f2d91fe4ac5b87ba6f9d`); `cmp` is silent and `diff -q` is empty; identical 1,000,012 lines / 191,891,037 bytes.
- The existing test suite passes **59/59** when run against the patched module.
- Edge cases verified byte-identical: interleaved ts, tied ts within a file and across files, malformed + non-object line skips, empty and single-record notebooks.
- Tie-break equivalence: for equal `ts`, `heapq.merge` yields the earlier iterable's items first; files are passed in `sorted(glob)` order with each file's equal-`ts` records contiguous, exactly reproducing Python's stable-sort order (file order, then line order).

## Trade-offs (stated honestly)

- **The win is headroom, not felt speed.** At lnb's intended scale (hundreds–low-thousands of entries) the old sort is already sub-millisecond — a typical user feels nothing. The value is (a) memory that scales to very large notebooks and (b) instant first output for piping into `jq`/`head`.
- **New assumption: each writer file must be `ts`-monotonic.** `heapq.merge` yields globally sorted output only if every input is already sorted. This holds by construction for append-only, single-writer-per-file notebooks, and was verified here (0 out-of-order in 1,000,012 records). Unlike the old `sorted()` — which globally re-sorts regardless — the merge is **not** robust to a file that is internally out-of-order (e.g. a backward clock step, a writer id shared across clock-skewed hosts, or a hand-edited file). In that case `log` emits that file's records in file order rather than globally sorted — deterministic, not a crash, and a documented assumption the code does not enforce. A cheap monotonicity guard was deliberately omitted because it would reintroduce a full pass and erase the streaming win; `scan_stream`'s docstring records the assumption.
- **+~40 lines, two read paths.** `log` uses `scan_stream`; `retract` keeps `scan`. Slightly more surface than a single sort call, in exchange for the memory/latency profile above.

## Scope

Limited to the `log` read path, per design. `scan()` (used by `retract`) is refactored only to share a per-file parse helper (`_iter_records`) with `scan_stream`; its `(records, retracted)` contract is unchanged.

https://claude.ai/code/session_01YUk9uiVCA75F2yTxo4Rh2x

### DIFF
--- a/lnb.py
+++ b/lnb.py
@@ -45,6 +45,8 @@ Discovery: $LNB_DIR, else the nearest `.lnb/` walking up from the cwd, else
 """
 import difflib
 import glob
+import heapq
+import itertools
 import json
 import os
 import re
@@ -105,6 +107,32 @@ def default_context():
 
 # --- the single read path & the single write path ---------------------------
 
+def _iter_records(path):
+    """Yield each record (a dict) from one .lnb/*.jsonl file, verbatim, skipping
+    blank lines. Fails closed per line -- a malformed line or a valid-JSON-but-
+    non-object (scalar/array) line is skipped with a warning, never crashes a
+    read. Shared by scan() and scan_stream() so both read a file identically.
+    """
+    try:
+        with open(path, encoding="utf-8") as fh:
+            for lineno, line in enumerate(fh, 1):
+                line = line.strip()
+                if not line:
+                    continue
+                try:
+                    rec = json.loads(line)
+                except json.JSONDecodeError:
+                    warn(f"skipping malformed line {os.path.basename(path)}:{lineno}")
+                    continue
+                if not isinstance(rec, dict):              # valid JSON, but a
+                    warn(f"skipping non-object line "      # scalar/array is not
+                         f"{os.path.basename(path)}:{lineno}")  # a record -- fail
+                    continue                               # closed, never crash
+                yield rec                                  # every record, verbatim
+    except OSError as e:
+        warn(f"cannot read {path}: {e}")
+
+
 def scan(nbdir):
     """Scan every .lnb/*.jsonl -> (all records incl. `_retract` tombstones,
     ascending by ts string; the retracted id set).
@@ -118,28 +146,28 @@ def scan(nbdir):
     """
     records, retracted = [], set()
     for path in sorted(glob.glob(os.path.join(nbdir, "*.jsonl"))):
-        try:
-            with open(path, encoding="utf-8") as fh:
-                for lineno, line in enumerate(fh, 1):
-                    line = line.strip()
-                    if not line:
-                        continue
-                    try:
-                        rec = json.loads(line)
-                    except json.JSONDecodeError:
-                        warn(f"skipping malformed line {os.path.basename(path)}:{lineno}")
-                        continue
-                    if not isinstance(rec, dict):              # valid JSON, but a
-                        warn(f"skipping non-object line "      # scalar/array is not
-                             f"{os.path.basename(path)}:{lineno}")  # a record -- fail
-                        continue                               # closed, never crash
-                    records.append(rec)                        # every record, verbatim
-                    if rec.get("type") == "_retract" and rec.get("retracts"):
-                        retracted.add(rec["retracts"])
-        except OSError as e:
-            warn(f"cannot read {path}: {e}")
+        for rec in _iter_records(path):
+            records.append(rec)                            # every record, verbatim
+            if rec.get("type") == "_retract" and rec.get("retracts"):
+                retracted.add(rec["retracts"])
     records = sorted(records, key=lambda e: e.get("ts", ""))    # ascending, no reverse=
     return records, retracted
+
+
+def scan_stream(nbdir):
+    """Stream all records ascending by ts via a k-way merge over the per-writer
+    files. Each file is already ts-ascending (append-only), so heapq.merge yields
+    a globally sorted stream in O(k) memory -- k open files -- vs scan()'s O(n)
+    load-and-sort. For records with equal ts, heapq.merge yields the earlier
+    iterable's items first (tie broken by iterable index); passing files in
+    sorted(glob) order, each with its equal-ts records contiguous, reproduces
+    scan()'s stable-sort tie-break (file order, then line order) exactly. Used
+    only by `log`; `retract` still uses scan() for the materialized set it needs.
+    """
+    return heapq.merge(
+        *(_iter_records(p) for p in sorted(glob.glob(os.path.join(nbdir, "*.jsonl")))),
+        key=lambda e: e.get("ts", ""),
+    )
 
 
 def append(nbdir, record):
@@ -232,12 +260,13 @@ def cmd_log(args):
         print('no notebook found here. Start one with:  lnb note "..."',
               file=sys.stderr)
         return
-    records, _ = scan(nbdir)
-    if not records:
+    stream = scan_stream(nbdir)                 # O(k) k-way merge, not a full load
+    first = next(stream, None)                  # peek: empty notebook -> nudge
+    if first is None:
         print(f'notebook at {nbdir} is empty -- nothing logged yet.\n'
               f'Start with:  lnb note "..."', file=sys.stderr)
         return
-    emit_json(records)
+    emit_json(itertools.chain([first], stream))  # emit_json iterates lazily
 
 
 def cmd_retract(args):
--- a/lnb.py
+++ b/lnb.py
@@ -39,6 +39,10 @@ of your own -- lnb ships none by design.
 
 `log` sorts on the ISO ts STRING and assumes a stable UTC offset (a single-
 timezone notebook); cross-offset instant ordering is a documented limitation.
+Because `log` now streams a k-way merge over the per-writer files, it also
+assumes each file is already ts-ascending -- true for append-only single-writer
+files; a backward clock step or a hand-edit can violate it, and then that file's
+records emit in file order rather than globally sorted.
 
 Discovery: $LNB_DIR, else the nearest `.lnb/` walking up from the cwd, else
 `./.lnb` is created on the first `note`. Writer: $LNB_WRITER, else $USER.
@@ -54,6 +58,11 @@ import secrets
 import subprocess
 import sys
 from datetime import datetime
+
+try:
+    import resource               # POSIX-only; sizes scan_stream's fan-out
+except ImportError:               # pragma: no cover - non-POSIX
+    resource = None
 
 CORE = ("id", "ts", "writer", "context", "type", "content")  # lnb sets these
 RESERVED = set(CORE) | {"retracts", "reason"}                # writers may not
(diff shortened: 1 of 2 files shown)


lnb.py
```diff
@@ -232,12 +260,13 @@ def cmd_log(args):
         print('no notebook found here. Start one with:  lnb note "..."',
               file=sys.stderr)
         return
-    records, _ = scan(nbdir)
-    if not records:
+    stream = scan_stream(nbdir)                 # O(k) k-way merge, not a full load
+    first = next(stream, None)                  # peek: empty notebook -> nudge
```
carbonscott, line 308: Nice: peeking with `next(stream, None)` and re-prepending via `itertools.chain([first], stream)` preserves the empty-notebook nudge exactly. Using the `None` sentinel (rather than a truthiness check) is the correct call here — a valid record that happens to be an empty dict `{}` is falsy, and a `not first` test would have misfired on it. This reads cleanly.

lnb.py
```diff
@@ -118,30 +146,30 @@ def scan(nbdir):
...
+    a globally sorted stream in O(k) memory -- k open files -- vs scan()'s O(n)
+    load-and-sort. For records with equal ts, heapq.merge yields the earlier
+    iterable's items first (tie broken by iterable index); passing files in
+    sorted(glob) order, each with its equal-ts records contiguous, reproduces
+    scan()'s stable-sort tie-break (file order, then line order) exactly. Used
+    only by `log`; `retract` still uses scan() for the materialized set it needs.
+    """
+    return heapq.merge(
+        *(_iter_records(p) for p in sorted(glob.glob(os.path.join(nbdir, "*.jsonl")))),
+        key=lambda e: e.get("ts", ""),
```
carbonscott, line 213: Consider: streaming makes ordering-time exceptions produce *partial* stdout. If the sort key raises mid-merge — e.g. a hand-edited/corrupt file with a non-string `ts`, which makes `heapq` compare `int` vs `str` — `log` emits some valid lines and *then* crashes with a `TypeError` traceback (I reproduced 2 lines emitted, then exit 1). The old `sorted()` raised the same error but before any output, so it failed closed with a clean, empty stdout. A near-free hardening that also keeps the two read paths identical: coerce the key in *both* `scan` and `scan_stream` to `str(e.get("ts", ""))`. For well-formed (all-string ts) input the output is unchanged, but it makes the ordering total and removes the mid-stream `TypeError` entirely. Non-blocking — the trigger is out-of-contract input — but worth a moment given the codebase prizes fail-closed reads.

lnb.py
```diff
@@ -118,30 +146,30 @@ def scan(nbdir):
...
     return records, retracted
 
 
+def scan_stream(nbdir):
+    """Stream all records ascending by ts via a k-way merge over the per-writer
+    files. Each file is already ts-ascending (append-only), so heapq.merge yields
+    a globally sorted stream in O(k) memory -- k open files -- vs scan()'s O(n)
+    load-and-sort. For records with equal ts, heapq.merge yields the earlier
+    iterable's items first (tie broken by iterable index); passing files in
+    sorted(glob) order, each with its equal-ts records contiguous, reproduces
```
carbonscott, line 196: Consider adding a test for this tie-break. The docstring (rightly) promotes equal-`ts` ordering — file order, then line order — to a documented contract, and I confirmed it currently matches stable `sorted()`. But no test pins it: `test_log_emits_all_ascending_across_writers` uses all-distinct timestamps, so a future change (dropping `sorted(glob)`, swapping the key, an unstable merge) could break the guarantee with the suite still green. A two-file, identical-`ts` case asserting the exact id order would lock the contract down cheaply.

lnb.py
```diff
@@ -118,30 +146,30 @@ def scan(nbdir):
...
+            records.append(rec)                            # every record, verbatim
+            if rec.get("type") == "_retract" and rec.get("retracts"):
+                retracted.add(rec["retracts"])
     records = sorted(records, key=lambda e: e.get("ts", ""))    # ascending, no reverse=
     return records, retracted
 
 
+def scan_stream(nbdir):
+    """Stream all records ascending by ts via a k-way merge over the per-writer
+    files. Each file is already ts-ascending (append-only), so heapq.merge yields
```
carbonscott, line 192: FYI / Consider: the per-file ts-monotonicity assumption is documented well here, but only in this internal docstring. The module-level docstring already spells out the ts-string / single-timezone limitation for `log`; adding one clause there about the per-file monotonic-append assumption would surface it where a user actually looks. Worth flagging one real, non-corruption trigger: a backward wall-clock step (e.g. an NTP correction) between two `note`s from the same writer yields a genuinely non-monotonic file with no hand-editing — in which case `log` silently emits that file in append order rather than globally sorted. The trade-off is reasonable; this is just about where the caveat is visible.

lnb.py
```diff
@@ -232,12 +260,13 @@ def cmd_log(args):
...
-    records, _ = scan(nbdir)
-    if not records:
+    stream = scan_stream(nbdir)                 # O(k) k-way merge, not a full load
+    first = next(stream, None)                  # peek: empty notebook -> nudge
+    if first is None:
         print(f'notebook at {nbdir} is empty -- nothing logged yet.\n'
               f'Start with:  lnb note "..."', file=sys.stderr)
         return
-    emit_json(records)
+    emit_json(itertools.chain([first], stream))  # emit_json iterates lazily
```
carbonscott, line 313: FYI: `scan_stream` returns a lazy generator that keeps k file handles open for the life of the merge, and CPython's `heapq.merge` does not close its sub-iterators on early `GeneratorExit` (no try/finally in its impl) — deterministic close relies on full consumption or process exit. For the one-shot `log` CLI this is a non-issue (I verified `lnb log | head -2` exits 0 with no traceback and no leak). Just noting the contract in case `scan_stream` is ever reused in a long-lived process: consume it fully, or the handles linger until GC.

lnb.py
```diff
@@ -118,30 +146,30 @@ def scan(nbdir):
...
+    files. Each file is already ts-ascending (append-only), so heapq.merge yields
+    a globally sorted stream in O(k) memory -- k open files -- vs scan()'s O(n)
+    load-and-sort. For records with equal ts, heapq.merge yields the earlier
+    iterable's items first (tie broken by iterable index); passing files in
+    sorted(glob) order, each with its equal-ts records contiguous, reproduces
+    scan()'s stable-sort tie-break (file order, then line order) exactly. Used
+    only by `log`; `retract` still uses scan() for the materialized set it needs.
+    """
+    return heapq.merge(
+        *(_iter_records(p) for p in sorted(glob.glob(os.path.join(nbdir, "*.jsonl")))),
```
carbonscott: Nit (optional): `sorted(glob.glob(os.path.join(nbdir, "*.jsonl")))` is now duplicated verbatim in `scan()` and `scan_stream()`. A tiny `_writer_files(nbdir)` helper would keep the two read paths' file discovery in lockstep (and document that sorted-glob order is load-bearing for the tie-break). Borderline — two one-liners may not clear the bar for a new helper — so purely your call.

carbonscott (commented):
> **Intended disposition: APPROVE.** GitHub forbids formally approving your own PR (the API returned *"Can not approve your own pull request"*), so this is posted as a COMMENT review. Substantively it is an approval: no blocking issues, all findings below are Consider/FYI/Nit.

## Review Summary

**Overall:** A tightly-scoped, well-executed performance change that swaps `log`'s materialize-then-`sorted()` read path for a streaming `heapq.merge` k-way merge. The peek-then-stream logic is correct, the tie-break is empirically byte-identical to the old stable sort, the `_iter_records` extraction is a clean de-duplication, and the existing 59-test suite passes (verified locally). I ran the multi-file merge, equal-ts tie-break, and early-termination paths by hand and all behave as documented. Approving. The comments below are all non-blocking (Consider/FYI/Nit) plus a couple of acknowledgements.

### Key Findings

**Functionality (correct):**
- `cmd_log`'s peek-then-chain is right: `next(stream, None)` uses the `None` sentinel (not truthiness), so the empty-notebook nudge fires only on a genuinely empty stream, and even a legitimate empty-dict `{}` record would be handled correctly. `itertools.chain([first], stream)` faithfully re-prepends the peeked record. Verified against `test_empty_notebook_diagnostic`.
- The equal-`ts` tie-break claim holds: I confirmed `heapq.merge` yields `[A1, A2, B1]` for identical `ts` across two files, identical to stable `sorted()` (file order, then line order).

**Functionality (behavioral difference worth noting, non-blocking):**
- Because ordering is now interleaved with emission, an exception raised *during ordering* can emit a **partial, truncated JSONL stream** before crashing, whereas the old `sorted()` failed atomically with zero stdout. I reproduced this: a corrupt file with a non-string `ts` surfacing mid-stream emitted 2 valid lines, then a `TypeError` traceback and exit 1. The trigger (mixed-type `ts`) is out-of-contract corrupt/hand-edited input — the same population as the documented monotonicity caveat — so severity is low, but it hands a downstream `jq` a torn stream and slightly undercuts the codebase's "fail closed" framing. See the inline note on the `key=` line for a cheap optional hardening.

**Tests:**
- Core coverage is good — `test_log_emits_all_ascending_across_writers` directly exercises the k-way merge across two interleaved writer files, and the empty/malformed/non-object paths are all driven through `log`. The one gap: the equal-`ts` tie-break, which this PR *elevates to a documented contract* in `scan_stream`'s docstring, has no dedicated test (the multi-writer test uses all-distinct ts). A future refactor could silently break it. Inline suggestion to add a same-ts-across-two-files pin.

**Documentation / Design:**
- The per-file monotonicity assumption is documented honestly and thoroughly, with an explicit rationale for omitting a guard. One consideration: it lives only in `scan_stream`'s internal docstring; a one-clause mention in the module-level docstring (which already documents the ts-string/single-timezone limitation) would surface it to users. Also note a real, non-corruption trigger: a backward wall-clock/NTP step between two `note`s by the same writer produces a non-monotonic file with no hand-editing.

### PR Description Feedback
Exemplary. Clear *what*/*why*, measured benchmarks with stated method, correctness proven three independent ways, an honest trade-offs section, and explicit scope boundaries. This is a model PR description — no changes requested.

6 inline comments left (all non-blocking).

PR #36, opened 2026-07-07
docs(examples): retrieval cookbook for reading .lnb JSONL without `lnb log`

carbonscott:
## What

A minimal, verified cookbook — `examples/retrieval.md` — of recipes for retrieving an lnb notebook's JSONL **directly from `.lnb/*.jsonl`, without `lnb log`**: `grep` point-lookups, a `grep`-as-pre-filter accelerator, `jq` projection / live-view / most-recent-N & time-range, and an optional throwaway SQLite `jsonb`+index for read-heavy bursts. Plus a single one-line pointer to it in `README.md` under "The honest tradeoff".

## Why

The README already frames the escalation path ("for the full-power case, `jq` directly over `.lnb/*.jsonl`") but doesn't show how. This cookbook is that missing page.

## Scope

Docs/examples only. `git diff --name-only main...HEAD` = `README.md` + `examples/retrieval.md`. `lnb.py` and all other tracked source are unchanged; no new dependencies (coreutils / jq / sqlite3 / Python stdlib only). The SQLite index is framed as optional, disposable, git-ignored, and never authoritative.

## Verification

Every fenced command block was executed against a real lnb notebook and returns the output shown. The two footguns of leaving `lnb log` — files not globally `ts`-sorted, and naive filters being fail-open on retractions — are called out up front, and the live-view recipe demonstrates the fix (11 raw records → 9 live). Curation: 6 recipes kept, each a distinct retrieval need; redundant / README-overlapping candidates cut (roster in the cookbook's design notes).

https://claude.ai/code/session_01GXQZGVjHLj7yPrKh6LSCzF

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,3 +99,5 @@ Every read is O(all entries) and nothing validates fields — fine at 10^4 entri
 a lie at 10^7. The escalation path when a scan outgrows the terminal is not a
 second query language; it is `lnb log | jq '…'`, and for the full-power case,
 `jq` directly over `.lnb/*.jsonl`.
+
+For a cookbook of those direct-access recipes — `grep` point-lookups, `jq` projections and filters, and an optional throwaway SQLite index for read-heavy bursts — see [examples/retrieval.md](examples/retrieval.md).
--- a/examples/retrieval.md
+++ b/examples/retrieval.md
@@ -83,13 +83,15 @@ grep -h -F 'AUROC' .lnb/*.jsonl | jq -c 'select(.type=="observation")'
 {"id":"20260707T211152-9ceb94eb","ts":"2026-07-07T21:11:52+00:00","writer":"cong","context":"ssl/eval","type":"observation","content":"baseline AUROC 0.91 on holdout","metric":"auroc","value":"0.91"}
 ```
 
-The speedup is real: on a 1,000,016-record / 191 MB notebook, one id lookup ran in
-**0.04 s** as `grep -F … | jq` versus **2.2 s** for a bare `jq 'select(.id==…)'`
-full scan — about **55× faster**, because `grep` never builds a parse tree for the
-999,999 lines it discards. (`grep` alone can only match raw bytes; the moment you
-need field logic — a specific `.type`, a comparison — `jq` still does that part.
-For a literal id, `grep` alone from *Point-lookup by id* is enough.) If you are
-doing *many* lookups rather than one, build the index under *Read-heavy bursts*.
+The speedup is real: on a 1,000,016-record / 191 MB notebook, this same shape —
+`grep -F '<phrase>' | jq 'select(.type==…)'` for a phrase occurring in a single
+record — ran in **0.06 s** versus **~2.5 s** for the equivalent bare
+`jq 'select(.content|contains("<phrase>"))'` full scan: about **40× faster**,
+because `grep` discarded all but the one matching line before `jq` parsed anything.
+(`grep` alone can only match raw bytes; the moment you need field logic — a specific
+`.type`, a comparison — `jq` still does that part. For a literal id, `grep` alone
+from *Point-lookup by id* is enough.) If you are doing *many* lookups rather than
+one, build the index under *Read-heavy bursts*.
 
 ## Project fields to a table — `jq`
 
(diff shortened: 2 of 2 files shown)


carbonscott (commented):
> Verdict: **LGTM** — reviewed with the /review-pr skill. Posted as a COMMENT because GitHub disallows a PR author formally Approving their own PR (and this session's guardrail blocks self-approval); `main` is unprotected, so no approval gate blocks merge.

## Review Summary

**Overall:** A clean, genuinely useful docs-only PR that fills a real gap — the README already names `jq` directly over `.lnb/*.jsonl` as the escalation path but never showed how; this cookbook is that missing page.

### Key Findings
- **Design / scope (strong):** Docs-only — `git diff --name-only main...HEAD` is exactly `README.md` + `examples/retrieval.md`; `lnb.py` and all other source are untouched, no new dependencies (coreutils / jq / sqlite3 / Python stdlib). The optional SQLite index is framed unambiguously as disposable, git-ignored, and never authoritative, so it does not undercut the project's deliberate no-built-in-index decision.
- **Functionality (strong):** Every fenced command block was executed against a real notebook and reproduces the shown output, including the `\$`-escaped SQLite lookups and `EXPLAIN QUERY PLAN` (`USING INDEX idx_id`). The two footguns of leaving `lnb log` — files not globally `ts`-sorted, and naive filters being fail-open on retractions — are stated up front and demonstrated (live-view goes 11 raw -> 9 live).
- **Complexity / minimality:** Six recipes, each a distinct retrieval need; redundant or README-overlapping candidates (artifacts split, plain type filter, content regex, aggregation) were deliberately cut. The two grep recipes are cleanly demarcated (point-lookup vs. pre-filter accelerator).
- **Portability (addressed in review):** `jsonb()` requires sqlite3 >= 3.45.0; the PR now calls this out with a verified `JSON`/`json()` text-storage fallback for older sqlite3.
- **Numbers (addressed in review):** the grep-accelerator figure was re-measured with the recipe's actual phrase+field pattern (~40x), and the SQLite break-even is now stated against both baselines (<1 lookup vs `lnb log | jq`, ~3 vs a bare `jq` scan).

### PR Description Feedback
- Good: clear What / Why / Scope / Verification; no action needed.

3 inline comments (2 acknowledgements, 1 optional). No blocking issues; merge once CI is green.

examples/retrieval.md
```diff
@@ -0,0 +1,280 @@
...
+20260707T211152-f53d2780	question	why does val loss plateau at epoch 40?
+20260707T211152-9ceb94eb	observation	baseline AUROC 0.91 on holdout
+20260707T211213-7b0d9416	dead-end	learning rate 3e-4 diverges after 2 epochs
+20260707T211213-a278c01f	_retract	
+```
+
+This ordering matches `lnb log` exactly. Without `-s | sort_by(.ts)`, ana's
+`21:11:52` rows print before cong's `21:11:51` row — backwards in time.
+
+## Filter to what's current — `jq`
```
carbonscott, line 123: Good call making this the correctness anchor. Reproducing the live-view idiom over the raw files (not just pointing at `lnb log`) is exactly what a reader who has *left* `lnb log` needs, and the 2-dead-ends -> 1 contrast makes the fail-open footgun concrete rather than abstract.

examples/retrieval.md
```diff
@@ -0,0 +1,280 @@
...
+This database is **not a second source of truth.** The `.lnb/*.jsonl` files stay
+authoritative. The DB is derived (built entirely from them), rebuildable (delete it
+and re-run), git-ignored (add `*.db` to `.gitignore`), and safe to `rm` at any
+time — `lnb` never reads it, writes it, or knows it exists. Put it *outside* `.lnb/`.
+
+Load each JSONL line into a `jsonb` column (binary JSON — smaller and faster to
+decode than text `json`) and index only the fields you'll seek on. A tiny
+stdlib-only loader is the most robust way to get raw lines in:
+
+> **Prerequisite:** `jsonb()` needs `sqlite3` ≥ 3.45.0 (Jan 2024) — check with
```
carbonscott, line 212: FYI: nice — gating this on `sqlite3 >= 3.45.0` with a `JSON`/`json()` fallback keeps the recipe honest for readers on Ubuntu 22.04 / Debian 12 / stock macOS, where `jsonb()` would otherwise fail with a bare `no such function` error.

README.md
```diff
@@ -99,3 +99,5 @@ Every read is O(all entries) and nothing validates fields — fine at 10^4 entri
 a lie at 10^7. The escalation path when a scan outgrows the terminal is not a
 second query language; it is `lnb log | jq '…'`, and for the full-power case,
 `jq` directly over `.lnb/*.jsonl`.
+
+For a cookbook of those direct-access recipes — `grep` point-lookups, `jq` projections and filters, and an optional throwaway SQLite index for read-heavy bursts — see [examples/retrieval.md](examples/retrieval.md).
```
carbonscott, line 103: Optional: this pointer runs a bit long (a compound sentence with an em-dash aside). It reads fine and satisfies the one-line-pointer intent, so no need to change — only trim if you want it terser.

